PR #1, opened 2016-12-14
Update pluggable_rule_engine.md

mcoonen:
We ran in trouble with using delayed rules larger than 2700 characters. 
I found the solution in a slide made by Jason Coposky (http://slides.com/irods/debugging-rules-and-microservices-17#/4).

Since this was nowhere to be found in the iRODS documentation, I propose this addition.


trel:
Thanks!

PR #4, opened 2017-01-25
Update irods_rule_language.md

xu-hao:
Delete inaccurate c syntax highlighting, which causes problem that text after "*" are sometimes as comments.


trel:
This renders as intended for mkdocs.

I have updated a couple places at the bottom in 63ef3e2cf8df320ef567f7f0b1e2e7be8ff20cae

PR #15, opened 2018-04-10
link to levels table

mcast:
Fixes irods/irods#3112

Found this old issue while checking something, hope this little fix is useful


trel:
perfect - thanks!

PR #26, opened 2019-05-26
correct names of api PEPS for put,get,copy

d-w-moore:
As recently as 4.2.5 (https://docs.irods.org/4.2.5/plugins/dynamic_policy_enforcement_points/#available-dynamic-peps) irods documentation online has misspelled the pre- and post-PEPs for data_obj GET, PUT and COPY operations. This is a fix for that document.

### DIFF
--- a/dynamic_peps.json
+++ b/dynamic_peps.json
@@ -1768,7 +1768,7 @@
 				"paramName": "collOprStat"
 			}]
 		}, {
-			"funcName": "api_data_copy",
+			"funcName": "api_data_obj_copy",
 			"paramList": [{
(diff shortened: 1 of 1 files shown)


trel:
That list is/was generated from walking the code.

I think they are correct as is - and have different signatures than the ones with 'data_obj' in their name.

All three mentioned in this PR are also elsewhere on the page, with a different signature.

d-w-moore:
Ok, misunderstood then, I see now there are (for example) both pep_api_data_put_pre and pep_api_data_obj_put_pre.

PR #46, opened 2020-04-20
_finally and migration from static peps

trel:


docs/getting_started/upgrading.md
```diff
@@ -53,6 +53,196 @@ It skips any rows with an "invalid" `resc_id` value. A `resc_id` is considered i
...
+Apr 19 14:56:44 pid:22819 NOTICE: writeLine: inString = tempZone
+Apr 19 14:56:44 pid:22819 NOTICE: writeLine: inString = 12
+```
+
+### pep_api_data_obj_put_post()
+
+Moving to the dynamic PEP that fires after a data object is put via the iRODS API
+```
+pep_api_data_obj_put_post(*INSTANCE_NAME, *COMM, *DATAOBJINP, *BUFFER, *PORTAL_OPR_OUT){
+    writeLine("serverLog",*DATAOBJINP)
```
jasoncoposky, line 90: Should we also demonstrate the output from *COMM to show user related information?

trel: There is a *COMM example farther down...

https://github.com/irods/irods_docs/pull/46/files#diff-b0492b86f9fe84c51f7c359012ff8f25R232

PR #57, opened 2020-07-02
[irods/irods#4084] Update Default Resource section (master)

alanking:
Default Resource Configuration now always returns SYS_RESC_DOES_NOT_EXIST.
Previously, an iput targeting a resource that does not exist would redirect the put
to the default resource if "null" was used in msiSetDefaultResc.


trel:
please #

alanking:
Done

PR #59, opened 2020-07-27
[irods/irods#5017] Add docs for Federation parallel transfer limitation (master)

alanking:


docs/system_overview/federation.md
```diff
@@ -223,3 +223,55 @@ index 05e3553..53cc440 100644
...
+Here are two sample implementations for detecting remote put (the logic is very similar for gets):
+
+#### Native Rule Engine Plugin
+```sh
+path_is_in_local_zone(*logical_path) {
+    *logical_path like "/"++$rodsZoneProxy++"/*"; # return true or false
+}
+
+pep_api_data_obj_put_post(*INSTANCE_NAME, *COMM, *DATAOBJINP, *BUFFER, *PORTAL_OPR_OUT) {
+    if (!path_is_in_local_zone(*DATAOBJINP.obj_path)) {
```
trel, line 246: why flip with the `!`?
just have local policy be first to keep the logic easier to read?     and possibly not even have a remote section?

alanking: Depends on what we want to convey, I guess. I just wanted to show that you could put anything there (with nothing being an option) and was unsure if it was possible to return early in an NREP rule. 

Side note: I consider Federation the "special case" here and it is just a personal preference to put the special case as the conditional and the alternative/normal case being the `else`. Happy to flip these around, though!

trel: fair enough on the special case as a reason.    i was thinking that since policy probably *would* continue to exist on a local zone, and maybe not at all on a remote zone... we'd show where they can put their local code... and then the programmer would work out how to do the other case.

i think we flip it in this case - make it easier to read for those already probably in stress.

alanking: Done

docs/system_overview/federation.md
```diff
@@ -223,3 +223,51 @@ index 05e3553..53cc440 100644
...
+
+#### Native Rule Engine Plugin
+```sh
+path_is_in_local_zone(*obj_path) {
+    *obj_path like "/"++$rodsZoneProxy++"/*"; # return true or false
+}
+
+pep_api_data_obj_put_post(*INSTANCE_NAME, *COMM, *DATAOBJINP, *BUFFER, *PORTAL_OPR_OUT) {
+    *obj_path = *DATAOBJINP.obj_path;
+
```
trel, line 247: consider removing empty line here

docs/system_overview/federation.md
```diff
@@ -223,3 +223,51 @@ index 05e3553..53cc440 100644
...
+    # cannot access the same globals as NREP at this time
+    # access zone name from the local server configuration
+    server_config_path = '/etc/irods/server_config.json'
+    with open(server_config_path, 'r') as f:
+        zone_name = json.load(f)['zone_name']
+        return obj_path.startswith('/' + zone_name + '/')
+
+def pep_api_data_obj_put_post(rule_args, callback, rei):
+    obj_path = str(rule_args[2].objPath)
+
```
trel, line 268: consider removing empty line here

docs/system_overview/federation.md
```diff
@@ -223,3 +223,51 @@ index 05e3553..53cc440 100644
...
+If the put or get is going to or coming from the remote zone, any data object- or catalog-level operations performed as part of the post-PEP will reuse the connection to the remote zone. In the case of a parallel transfer, the put/get will return before any data is transferred because a separate portal operation will be performed between the client and server hosting the data. The post-PEP will fire on the local server to which the client is connected and will block returning to the client as it attempts to perform its operation. During this time, the remote server will time out and the portal will be shut down. The incoming request from the post-PEP will then be serviced, control returned to the local federating server, and then back to the client. The connection to the remote server will fail as the portal is no longer accepting connections.
+
+This is an architectural limitation between dynamic PEPs, federation, and the current parallel transfer implementation for puts and gets.
+
+Here are two sample implementations for detecting remote put (the logic is very similar for gets):
```
trel, line 237: "detecting a local zone put"

docs/system_overview/federation.md
alanking: Done

docs/system_overview/federation.md
alanking: Done

docs/system_overview/federation.md
alanking: Done

Issue #62, opened 2020-08-07
Add option to target different fork of repo


korydraughn:
- [x] master
- [x] 4-2-stable
---
This allows developers to verify the results of documentation in feature branches.

PR #63, opened 2020-08-07
[#62] Added support for building docs for different fork of repo (master)

korydraughn:


trel:
should we just pass in a git_repository... ?

and then it could be a full URL, or a local path... ?

trel:
MAKEIRODSREPOSITORY

korydraughn:
could it really be a local path?  supporting local repos would require more changes to make that work, right?

korydraughn:
do we want to add some underscores in the variable names? they are a bit hard to read.

Sticking with https://github/${GITHUB_ACCT}/irods somewhat guarantees that the repo being cloned is a fork of the irods/irods repo, right?

README.md
```diff
@@ -25,7 +26,7 @@ Build
 The `Makefile` default target will build two other targets, 'doxygen' and 'mkdocs'.
 
 ```
-$ make
+$ make [MAKEGITHUBACCOUNT=<value>] [MAKEIRODSVERSION=<value>]
```
trel, line 29: change this to the default value here?

PR #72, opened 2020-10-13
document error() and state() msvc

d-w-moore:


trel:
please cherry-pick to 4-2-stable as well

PR #105, opened 2021-06-04
Updated MySQL documentation (4-2-stable)

korydraughn:


docs/getting_started/installation.md
```diff
@@ -75,6 +75,12 @@ N rows in set (0.00 sec)
 
 ~~~
 
+!!! Note
+    MySQL v5.7 uses a default transaction isolation level of REPEATABLE-READ. You'll need to change this to READ-COMMITTED so that iRODS operates properly. This setting needs to be set globally. One way to achieve this is by starting MySQL with --transaction-isolation=READ-COMMITTED. For more information, see https://dev.mysql.com/doc/refman/5.7/en/innodb-transaction-isolation-levels.html.
+    
+!!! Note
+    MySQL v5.7 queries operate in a case-insensitive manner by default. iRODS requires that data be handled in a case-sensitive way. One way to achieve this is by starting MySQL with --character-set-server=utf8 and --collation-server=utf8_bin. However, due to the design of the iRODS database tables, the database may not start. As an alternative, you can set --collation-server=latin_general_cs (MySQL defaults to a latin1 character-set). For more information, see https://dev.mysql.com/doc/refman/5.7/en/charset-applications.html.
```
trel, line 82: use markdown links...  with the `[linktext](url)` notation... just keep the url as the linktext, that's fine.

and put backticks around the options in these two notes (like `--character-set-server=utf8`)

korydraughn:
This is the error message we get when trying to use utf-8.
```bash
irods.pypyodbc.Error: ('HY000', '[HY000] [MySQL][ODBC 5.2(a) Driver][mysqld-5.7.34]Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs')
```
I found documentation about the utf8 settings and restrictions. It may be worth mentioning the following articles in the notes section.
- https://dev.mysql.com/doc/refman/5.7/en/column-count-limit.html
- https://mariadb.com/kb/en/troubleshooting-row-size-too-large-errors-with-innodb/

MySQL and MariaDB appear to have ways around this, so it is likely up to the DBA to handle this. I'll try and play with these settings after UGM.

Thoughts?

Just an FYI, I learned that utf8 in MySQL land is not real utf8. What we want is the utf8mb4 variant.

trel:
yes, the multibyte.    it's a hot mess.

i'm fine linking to the mysql/mariadb docs - perhaps we show what the error would be if they don't consider it...  this part seems most google-able `Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535.`

korydraughn:
i tried to word things so we aren't responsible for bad stuff.

you think i should take the text out of the **Note** blocks? that would allow us to use lists and apply better structure.

korydraughn:
i think the current state of the PR is good enough right now.

i can circle back to this after UGM or after all of my slides are done if that works.

trel:
agreed - let's get it numbered and squashed...

docs/system_overview/configuration.md
```diff
@@ -33,6 +47,8 @@ This file contains the following top level entries:
 
     - `transfer_chunk_size_for_parallel_transfer_in_megabytes` (required) (default 40)
 
+  - `client_api_whitelist_policy` (default "") - If set to "enforce", the server will limit the scope of operations available to non-administrative clients.
```
trel, line 50: empty string is default?

korydraughn: I'm confused about what the context is around _default_. Is it for vanilla installs or when the option is not explicitly defined due to an upgrade?

trel: i think we're talking here about a vanilla installation.    upgrade (undefined) considerations are subtly different, agreed.

korydraughn: ok. i'll set it to "enforce"

korydraughn:
If we're happy with the wording, I'll add the #'s.

trel:
yes, if you've confirmed it renders as expected (indentation), please # em.

korydraughn:
added #

PR #121, opened 2022-01-14
[irods/irods#6089] Add comment about archive replicas and checksums (4-2-stable)

alanking:
Companion to irods/irods#6110


trel:
seems good - please # it

alanking:
#'d

Issue #20, opened 2018-11-19
Add developer documentation


alanking:
- [x] Deprecation/end of life (see #124)
- [x] Development environment (see https://github.com/irods/irods_docs/issues/219)
- [x] Building and installing iRODS (https://github.com/irods/irods_docs/issues/220)
- [x] Onboarding new developers (https://github.com/irods/irods_docs/issues/221)
- [x] Testing environment (https://github.com/irods/irods_docs/issues/222)
- [x] How to use Github
- [x] How to use rr/debugging tools (https://github.com/irods/irods_docs/issues/223)
- [x] Valgrind (https://github.com/irods/irods_docs/issues/224)
- [x] Static analysis (Clang) (https://github.com/irods/irods_docs/issues/225)
- [x] Running (and adding) tests (https://github.com/irods/irods_docs/issues/226)

alanking:
Could probably use some documentation around plugin operations as well as instructions for generic (and specific) plugin implementation.

korydraughn:
this should include documentation on coding conventions, for example, the JSON property name for instructing the server to execute the operation as an administrator should be *admin_mode* and accept a boolean value.

trel:
best practices section...
 - should include target library advice / do and do nots. 
     -  https://github.com/irods/irods/issues/6218#issuecomment-1056978356

korydraughn:
add notes to the developer docs about invoking external programs from C++ code.

PR #127, opened 2022-03-28
[#20] Update documentation for logging library (main)

korydraughn:
@trel What is the iteration process for this?

I made changes before standing everything up and those were visible in the browser. Since then, I've made additional changes but those don't appear at all. I've tried restarting the httpd server and that didn't work.

I feel like something has been cached.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -24,4 +24,4 @@ jobs:
               sudo apt-get install irods-icommands cmake cmake-data
       - name: Build
         run:  |
-              make
+              make BUILDAREA=/tmp DOXYGENMAKEJOBCOUNT=1
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,5 @@ RUN \
   mkdir -p /root/.irods && \
   echo "{}" | tee /root/.irods/irods_environment.json
 
-COPY . /irods_docs
-RUN ["chmod", "+x", "/irods_docs/run_make.sh"]
-ENTRYPOINT ["/irods_docs/run_make.sh"]
+COPY run_make.sh /run_make.sh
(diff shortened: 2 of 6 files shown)


trel:
The irods/irods code and the doxgen downloads are cached.

Any changes to docs should get put into the docker container if you rebuild it.

korydraughn:
When I make changes, how should I rebuild it?

I tried running `docker run --rm -v $(pwd)/build:/hostcomputer irods_docs_builder`, but it didn't work for me.

trel:
Re-run the build step.

This is the line that gets the new files into the container:
https://github.com/irods/irods_docs/blob/main/Dockerfile#L44

Happy to be told a better way... @alanking ?

korydraughn:
I think we have to pull that out of the image building step and provide irods_docs via a volume mount. It avoids producing several images due to the iteration process and results in an image that only contains the tools to build the docs.

I'll give that a try.

alanking (approved):
Looks good. I'll wait for input from @trel before proceeding.

I created #129 for the Github Actions failure.

korydraughn:
Sure. I may look into fixing #129 since I'm already here.

korydraughn:
Something is clearly wrong with the github actions when it tries to compile doxygen. The compiler seems to fail in a weird way.

korydraughn:
Turns out building doxygen used too many resources. Removing `-j` from the doxygen build step allowed the github action to work.

I've updated the project so that the value to `-j` is configurable.

korydraughn:
Any last thoughts/concerns before I add the #'s?

korydraughn:
Any last thoughts/concerns before I add the #'s?

trel:
seems good

PR #132, opened 2022-04-05
[#107] Updated configuration-specific documentation (main)

korydraughn:


docs/system_overview/configuration.md
```diff
@@ -86,30 +132,33 @@ This file contains the following top level entries:
...
 
+  - `zone_key` (required) - The shared secret used for authentication and identification of server-to-server communication - this can be a string of any length, excluding the use of hyphens, for historical purposes.  This must be the same across all iRODS servers in a Zone
+
+  - `zone_name` (required) - The name of the Zone in which the server participates. This must be the same across all iRODS servers in a Zone
+
+  - `zone_port` (required) (default 1247) - The main port used by the Zone for communication. This must be the same across all iRODS servers in a Zone
+
+  - `zone_user` (required) - The name of the rodsadmin user running this iRODS instance
 
 Servers in the catalog provider role have the following additional top level entries related to the database connection:
```
trel, line 164: these next ones are not top-level, they live in the database plugin section...

korydraughn: done

docs/system_overview/configuration.md
```diff
@@ -56,20 +68,54 @@ This file contains the following top level entries:
...
+
+  - `host_access_control` (required) - See [Host Access Control](#host-access-control) for more details
+
+      - `access_entries` (required)
+
+  - `host_resolution` (required) - See [Host Resolution](#host-resolution) for more details
+
+      - `host_entries` (required)
 
   - `kerberos_name` (optional) - Kerberos Distinguished Name for KRB and GSI authentication
```
korydraughn, line 91: Should this be listed here?

The name doesn't appear in the Kerberos configuration section. Given that there's a whole section on how to configure Kerberos, it seems reasonable to remove this.

docs/system_overview/configuration.md
```diff
@@ -82,132 +128,126 @@ This file contains the following top level entries:
 
   - `pam_password_min_time` (optional) - Minimum allowed PAM password lifetime, in seconds
 
-  - `re_data_variable_mapping_set` (required) - An array of file names comprising the list of data to variable mappings used by the rule engine, for example: [ "core" ] which references 'core.dvm'
+  - `plugin_configuration` (required) - Holds several different properties such as database connection information, rule engine configuration, etc.
 
-  - `re_function_name_mapping_set` (required) - An array of file names comprising the list of function name map used by the rule engine, for example: [ "core" ] which references 'core.fnm'
+      - `authentication` (optional) -
 
-  - `re_rulebase_set` (required) - An array of file names comprising the list of rule files used by the rule engine, for example: [ "core" ] which references 'core.re'.  This array is ordered as the order of the rule files affects which (multiply) matching rule would fire first.
```
korydraughn, line 89: Notice I removed the `re_*` properties.

These look like they are from a legacy version of iRODS. Aren't they part of the NREP now?

docs/system_overview/configuration.md
```diff
@@ -82,132 +128,126 @@ This file contains the following top level entries:
 
   - `pam_password_min_time` (optional) - Minimum allowed PAM password lifetime, in seconds
```
korydraughn, line 129: Should PAM be listed here?

Seems like the section describing how to setup PAM would handle these. Same for Kerberos.

korydraughn (commented):
Based on the comments, maybe we should list only the options that users have access to out-of-the-box.

Also, the use of `(required)` and `(optional)` seem inconsistent. Optional is implicit if `(required)` is attached to certain properties. I haven't counted, but we can invert their use if most properties are considered required.

docs/system_overview/configuration.md
trel: I think that's a correct assessment - it does feel weird to see it there now.

docs/system_overview/configuration.md
trel: yes, they are now plugin_specific_configuration....

trel:
we can consider if making a table, rather than a bulleted list is easier to read...

columns for 'name', 'required', 'default', and 'description'?

trel:
one of the things a table would lose is conveying 'nested-ness'.

korydraughn:
I like the idea of a table, but i also like the nested-ness.

Hmm ... I'll think on it, maybe even try it to see what it would look like.

docs/system_overview/configuration.md
```diff
@@ -14,65 +14,305 @@ This scrambled password file is saved after an `iinit` is run. If this file does
 
 This file defines the behavior of the server Agent that answers individual requests coming into iRODS. It is created and populated by the installer package.
 
-This file contains the following top level entries:
-
-  - `advanced_settings` (required) - Contains subtle network and password related variables.  These values should be changed only in concert with all connecting clients and other servers in the Zone.
-
-    - `default_log_rotation_in_days` (default 5) - The number of days the log is written to before being rotated out.
+!!! IMPORTANT
+    All configuration options are required unless marked as optional.
```
trel, line 18: don't want to call them options... perhaps... 'configuration settings'?

docs/system_overview/configuration.md
```diff
@@ -14,65 +14,305 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+    // Defines the filename of JSON schema used for validation when not specified explicitly.
+    "schema_name": "server_config",
 
-    - `default_temporary_password_lifetime_in_seconds` (required) (default 120) - The number of seconds a server-side temporary password is good.
+    // Defines the version of the schema used for validation.
+    // This normally maps to a directory under /var/lib/irods/configuration_schemas.
+    "schema_version": "v4",
+    
+    // Contains subtle network and password related variables.
+    // These values should be changed only in concert with all connecting clients and other servers in the Zone.
```
trel, line 30: password?

some of these settings are definitely this-server-only, so don't require 'in concert' edits elsewhere.

perhaps this note is mentioned on each setting where this is true.

docs/system_overview/configuration.md
```diff
@@ -14,65 +14,305 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+        "delay_rule_executors": [],
 
-    - `maximum_number_of_concurrent_rule_engine_server_processes` (optional) (default 4)
+        // The number of seconds the rule execution server sleeps before checking the delay queue for rules ready for execution.
+        "delay_server_sleep_time_in_seconds" : 30,
 
-    - `maximum_size_for_single_buffer_in_megabytes` (required) (default 32)
+        // Contains options for controlling the behavior of the internal DNS cache.
+        "dns_cache": {
+            // The maximum amount of shared memory allocated to the DNS cache.
```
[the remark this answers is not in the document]
korydraughn: it is pre-allocated at that size. the dns cache is not allowed to use more than that amount.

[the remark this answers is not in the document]
trel: right, so, just say 'amount of shared memory', no 'maximum'

docs/system_overview/configuration.md
```diff
@@ -14,65 +14,305 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
-    - `rule_engine_server_sleep_time_in_seconds` (optional) (default 30) - The number of seconds the rule execution server sleeps before checking the delay queue for rules ready for execution.
-    
-    - `transfer_buffer_size_for_parallel_transfer_in_megabytes` (required) (default 4)
+            // The amount of time an entry stays valid before being evicted from the cache.
+            "eviction_age_in_seconds": 3600
+        },
 
-    - `transfer_chunk_size_for_parallel_transfer_in_megabytes` (required) (default 40)
+        "hostname_cache": {
+            // The maximum amount of shared memory allocated to the Hostname cache.
```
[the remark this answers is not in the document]
trel: "The amount of shared memory", no maximum.   cool.

docs/system_overview/configuration.md
```diff
@@ -14,65 +14,305 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
-  - `default_dir_mode` (required) (default "0750") - The unix filesystem octal mode for a newly created directory within a resource vault
+        "maximum_size_for_single_buffer_in_megabytes": 32,
+
+        // The maximum number of bytes available to the delay queue.
+        // If set to 0, the delay server will use as much memory as it needs to hold queued rules.
+        "maximum_size_of_delay_queue_in_bytes": 0,
+
+        "maximum_temporary_password_lifetime_in_seconds": 1000,
+
+        // The number of delay rules allowed to execute simultaneously.
```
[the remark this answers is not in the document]
korydraughn: This applies to the delay server only.

It defines the number of worker threads used to execute delay rules. Each thread will process one rule at a time. So it is directly related to the number of agents the delay server could be interacting with simultaneously.

docs/system_overview/configuration.md
```diff
@@ -14,65 +14,305 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+        "stacktrace_file_processor_sleep_time_in_seconds": 10,
+
+        "transfer_buffer_size_for_parallel_transfer_in_megabytes": 4,
+
+        "transfer_chunk_size_for_parallel_transfer_in_megabytes": 40
+    },
+
+    // Controls which API operations are accessible to users.
+    // If set to "enforce", the server will expose a subset of operations that can be executed by any user. Administrative operations
+    // will require rodsadmin level privileges. This option does not apply to rodsadmin users.
```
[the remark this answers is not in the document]
korydraughn: We changed its type to be enum in the schema. it only accepts "enforce" or "disabled".

You can view it live right now to see the current draft.

docs/system_overview/configuration.md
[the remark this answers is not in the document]
trel: Okay, so this only has an effect if *this* machine is executing delay rules, aka, it's in the executors list.

Another machine in the executor list could have a different setting, and be pulling down its own N jobs.

The default case, for now, is that the local machine is the *only* executor, so, same as the machine designated as the delay_server.

docs/system_overview/configuration.md
```diff
@@ -14,244 +14,640 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+    "catalog_service_role": "",
+
+    // Controls which API operations are accessible to non-rodsadmin users.
+    //
+    // When set to "enforce", the server will expose a subset of operations that can be executed by any user.
+    // Administrative operations will require rodsadmin level privileges. This option does not apply to
+    // rodsadmin users.
+    //
+    // When set to "disabled", the server will operate without consulting the allowlist. This is equivalent
+    // to how server behaved prior to iRODS 4.3.0.
```
trel, line 102: - how "the" server behaved

docs/system_overview/configuration.md
```diff
@@ -14,244 +14,640 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+    //
+    // The following values are supported: enforce, disabled
```
trel, line 104: consider ...

> The following two values are supported:
> 
>  - When set to "enforce" ...
> 
>  - When set to "disabled" ...

docs/system_overview/configuration.md
```diff
@@ -14,244 +14,640 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+    // Only applies to service account environment files.
+    "schema_version": "v4"
+}
+```
+
+A client environment file might look like this.
+
+```json
+{
+    "irods_host": "node-1.example.company.com",
```
trel, line 613: let's stick to example.org as the root domain, since it's designed for this usage (documentation)

docs/system_overview/configuration.md
```diff
@@ -14,244 +14,640 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+    "irods_cwd": "/tempZone/home/rods",
+    "irods_default_hash_scheme": "SHA256",
+    "irods_default_number_of_transfer_threads": 4,
+    "irods_default_resource": "demoResc",
+    "irods_encryption_algorithm": "AES-256-CBC",
+    "irods_encryption_key_size": 32,
+    "irods_encryption_num_hash_rounds": 16,
+    "irods_encryption_salt_size": 8,
+    "irods_home": "/tempZone/home/rods",
+    "irods_host": "node-1.example.company.com",
```
trel, line 636: `node-1.example.org` here too

docs/system_overview/configuration.md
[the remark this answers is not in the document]
korydraughn: Not quite.

Yes, this only affects the machine running the delay server.

The executors list is different in that it defines the set of iRODS servers the delay server will dispatch rules to for execution. Those servers do not pull work from the catalog.

The delay server contains a thread pool. This settings represents the size of that pool.

docs/system_overview/configuration.md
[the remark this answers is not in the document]
korydraughn: done

docs/system_overview/configuration.md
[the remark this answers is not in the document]
korydraughn: done

docs/system_overview/configuration.md
korydraughn: done

docs/system_overview/configuration.md
korydraughn: done

docs/system_overview/configuration.md
korydraughn: done

docs/system_overview/configuration.md
korydraughn: removed

docs/system_overview/configuration.md
korydraughn: removed

docs/system_overview/configuration.md
korydraughn: Changed to the following to keep things inline with how other properties define what range of acceptable values.

Basically, I moved the line declaring the values above the lines describing what each value means.

I feel this helps the reader in two ways. (1) They are informed of the accepted values. (2) By listing the options up-front, they don't have to look over the entire list to see what is valid. They can pick and choose which one they are interested in. This should probably be the rule for all settings.
```
Controls which API operations are accessible to non-rodsadmin users.

The following values are supported: enforce, disabled

When set to "enforce", the server will expose a subset of operations that can be executed by any user.
Administrative operations will require rodsadmin level privileges. This option does not apply to rodsadmin users.

When set to "disabled", the server will operate without consulting the allowlist. This is equivalent
to how the server behaved prior to iRODS 4.3.0.
```

docs/system_overview/configuration.md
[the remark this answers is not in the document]
trel: Ah, okay.   So this setting is ignored / has no effect on any machine other than the active delay server.

Got it.

docs/system_overview/configuration.md
trel: Yeps, looks good.

docs/system_overview/configuration.md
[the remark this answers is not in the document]
korydraughn: Correct

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,676 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+            "shared_memory_size_in_bytes": 2500000,
 
-  - `db_password` (required) - The password for the `db_username` to connect to the `db_name`
+            // The amount of time an entry stays valid before being evicted from the cache.
+            "eviction_age_in_seconds": 3600
+        },
 
-  - `db_port` (required) - The port on which the database server is listening
+        // TODO
+        "maximum_size_for_single_buffer_in_megabytes": 32,
```
korydraughn, line 63: Anyone know what this is?

Is it only used for put operations?

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,676 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+
+        // (Optional)
+        // The maximum number of bytes available to the delay queue.
+        // When set to 0, the delay server will use as much memory as it needs to hold queued rules.
+        "maximum_size_of_delay_queue_in_bytes": 0,
+
+        // TODO
+        "maximum_temporary_password_lifetime_in_seconds": 1000,
```
korydraughn, line 71: Needs a description.

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,676 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+        // The number of delay rules allowed to execute simultaneously.
+        "number_of_concurrent_delay_rule_executors": 4,
+
+        // (Optional)
+        // The amount of time the stacktrace file processing thread sleeps before files are
+        // processed and written to the log file.
+        "stacktrace_file_processor_sleep_time_in_seconds": 10,
+
+        // TODO
+        "transfer_buffer_size_for_parallel_transfer_in_megabytes": 4,
```
korydraughn, line 83: Needs a description.

Is this specific to parallel transfer via a put operations?

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,676 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+
+        // TODO
+        "transfer_chunk_size_for_parallel_transfer_in_megabytes": 40
```
korydraughn, line 86: Needs a description.

Is this specific to parallel transfer via a put operation?

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,676 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+    "match_hash_policy": "compatible",
+
+    // A 32-byte encryption key shared by the zone for use in the advanced negotiation handshake at the
+    // beginning of an iRODS client connection.
+    "negotiation_key": "",
+
+    // Holds several different properties such as database connection information, rule engine configuration, etc.
+    "plugin_configuration": {
+        // TODO
+        "authentication": {},
```
korydraughn, line 245: Needs a description.

When is this used? `izonereport` perhaps? Anything else?

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,676 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
-  - `db_odbc_type` (required) - The ODBC type, usually 'unix'
-  - `db_password` (required) - The password for the `db_username` to connect to the `db_name`
-  - `db_port` (required) - The port on which the database server is listening
-  - `db_name` (required) - The name of the database used as the iCAT
-  - `db_username` (required) - The database user name
+        // TODO
+        "network": {},
+
+        // TODO
+        "resource": {},
```
korydraughn, line 278: Needs a description.

When is this used? izonereport perhaps? Anything else?

docs/system_overview/configuration.md
korydraughn: This doesn't appear to be used/set by anything.

We might be able to remove this.

docs/system_overview/configuration.md
trel: i think this is just for completeness - that each plugin type has a section.     i don't know that we use this section, since we have 'context string' per resource for most of this type of thing.

docs/system_overview/configuration.md
trel: not just put - any parallel transfer (put/get/repl).

docs/system_overview/configuration.md
trel: not just put - any single buffer transfers (put/get/repl).

docs/system_overview/configuration.md
korydraughn: I've marked these as "This property can be ignored."

docs/system_overview/configuration.md
korydraughn: Marked as "This property can be ignored."

korydraughn:
Any additional thoughts on this PR?

trel:
looking - will render it

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,682 @@ This scrambled password file is saved after an `iinit` is run. If this file does
 
 This file defines the behavior of the server Agent that answers individual requests coming into iRODS. It is created and populated by the installer package.
 
-This file contains the following top level entries:
+!!! IMPORTANT
+    All configuration settings are required unless marked as optional. The value associated with a key represents the default value.
```
alanking, line 18: I'm guessing these are the default values according to the schema?

korydraughn: All defaults come from the template files.

I defined the value as an empty string or zero if there is no default defined in the template file.

trel: not according to the schema - just the server code itself.     and therefore, perhaps, the template.

alanking: Ah, the templates. Thanks

docs/irods.css
```diff
@@ -94,3 +94,8 @@ a:hover {
 .rst-content .section ol p, .rst-content .section ul p {
     margin-bottom: 3px;
 }
+span.hljs-comment {
+    color: #6f99c3;
+    font-style: italic;
```
trel, line 99: i think this looks better with `normal`, rather than `italic`.

?

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,682 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+```json
+{
+    "irods_host": "example.org",
+    "irods_port": 1247,
+    "irods_user_name": "alice",
+    "irods_zone_name": "tempZone"
+}
 ```
+
+And here is service account environment file defining the minimum settings needed to manage an iRODS server.
```
alanking, line 652: here is service account environment file -> here is a service account environment file ?

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,682 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+
+```bash
 export IRODS_ENVIRONMENT_FILE=/full/path/to/different.json
 ```
 
 Other individual environment variables can be set by using the UPPERCASE versions of the variables named above, for example:
 
-```
+```bash
 export IRODS_LOG_LEVEL=7
```
alanking, line 693: This option is deprecated, so maybe we shouldn't use it as the example anymore.

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,682 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+    // the use of the default hash scheme: strict or compatible.
+    "match_hash_policy": "compatible",
+
+    // A 32-byte encryption key shared by the zone for use in the advanced negotiation handshake at the
+    // beginning of an iRODS client connection.
+    "negotiation_key": "",
+
+    // Holds several different properties such as database connection information, rule engine configuration, etc.
+    "plugin_configuration": {
+        // This property can be ignored.
```
trel, line 245: `// This property is currently unused.`

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,682 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+        },
 
-  - `catalog_database_type` (required) - The type of database iRODS is using for the iCAT, either 'postgres', 'mysql', or 'oracle'
-  - `db_host` (required) - The hostname of the database server (can be localhost)
-  - `db_odbc_type` (required) - The ODBC type, usually 'unix'
-  - `db_password` (required) - The password for the `db_username` to connect to the `db_name`
-  - `db_port` (required) - The port on which the database server is listening
-  - `db_name` (required) - The name of the database used as the iCAT
-  - `db_username` (required) - The database user name
+        // This property can be ignored.
```
trel, line 275: `// This property is currently unused.`

docs/system_overview/configuration.md
```diff
@@ -14,254 +14,682 @@ This scrambled password file is saved after an `iinit` is run. If this file does
...
+        "network": {},
+
+        // This property can be ignored.
```
trel, line 278: `// This property is currently unused.`

alanking (commented):
Just noticed a couple of things. Overall looks very good

docs/system_overview/configuration.md
trel: that seems reasonable... what is a good replacement.... most-often overridden besides log level?

alanking: Mmmm maybe `irods_cwd`/`IRODS_CWD`?

We don't have to replace this, just seemed like an accident to make it the example when we've marked it as deprecated. I feel like the environment variable override feature should be removed, personally, but maybe there's a use case for it that I'm not thinking of :)

alanking:
```bash
export IRODS_CWD='/tempZone/home/public'
```
That seems to work

trel: it's definitely a use case - people love some environment variables in their docker containers

docs/irods.css
korydraughn: This styling came from the tool that generated the CSS. I only changed the color.

Will change and post what it looks like. You can open the developer tools and play with the colors and things in realtime. Just an fyi.

korydraughn: Normal looks fine to me. Will use that.

docs/system_overview/configuration.md
korydraughn: I was wondering if we should be advertising `IRODS_LOG_LEVEL` at all, but then I remembered we still need it for some tests because we haven't fully removed everything related to the old logging code.

For that reason, we do need to at least document it for ourselves.

docs/system_overview/configuration.md
korydraughn: done

docs/system_overview/configuration.md
korydraughn: updated.

docs/irods.css
korydraughn: done

docs/system_overview/configuration.md
korydraughn: done

docs/system_overview/configuration.md
korydraughn: done

docs/system_overview/configuration.md
korydraughn: done

trel:
i think it's ready to be squashed.

korydraughn:
Squashed. Should I go ahead and add the pounds?

trel:
yes please

korydraughn:
Added #'s

korydraughn:
@alanking poke

alanking (approved):

PR #148, opened 2022-07-26
Policy Cookbook and Library Examples (main)

korydraughn:
- Introduces a new section for capturing known policy techniques to various problems
- Absorbs the https://github.com/irods/irods_library_examples

### DIFF
--- a/docs/administrators/policy_cookbook.md
+++ b/docs/administrators/policy_cookbook.md
@@ -1,0 +1,471 @@
+# The iRODS Policy Cookbook
+
+This section is all about presenting techniques to various policy-based situations encountered in the iRODS ecosystem.
+
+Over time, we hope this cookbook will grow and become a valuable resource for new and existing users of iRODS.
+
+If you have suggestions on how to improve the cookbook or you've developed solutions that you feel could help others, please reach out to us by creating an issue at [https://github.com/irods/irods_docs](https://github.com/irods/irods_docs).
+
+## Synchronizing Delay Rules using Metadata
+
+This example demonstrates how to synchronize actively running delay rules using metadata.
+
+### How to do it ...
+
+The rules in the example below do the following:
+
+- Sequentially launch _N_ number of delay rules
+- Each delay rule waits for a unique piece of metadata to be attached to a collection before doing work
+- Assume the user has permission to modify metadata on `/tempZone/home/rods`
+
+Pay close attention to the use of `wait_for_metadata_signal` and `msiModAVUMetadata`. They are what make this technique possible.
+
+```python
+# Returns true if "*attribute_name" is attached to "*collection" as metadata.
+# Returns false otherwise.
+is_metadata_attached_to_collection(*collection, *attribute_name)
+{
+    *attached = false;
+
+    foreach (*row in select COLL_NAME where COLL_NAME = '*collection' and META_COLL_ATTR_NAME = '*attribute_name') {
+        *attached = true;
+    }
+
+    *attached;
+}
+
+# A convenience rule that acts as a spin-lock.
+# This rule also allows us to write clear and maintainable policy.
+wait_for_metadata_signal(*collection, *attribute_name)
+{
+    while (!is_metadata_attached_to_collection(*collection, *attribute_name)) {
+        msiSleep('1', '0');
+    }
+}
+
+launch_synchronized_delay_rules_example
+{
+    *number_of_delay_rules = 5;
+
+    # Schedule multiple delay rules.
+    for (*i = 0; *i < *number_of_delay_rules; *i = *i + 1) {
+        *signum = *i;
+
+        delay('<INST_NAME>irods_rule_engine_plugin-irods_rule_language-instance</INST_NAME>') {
+            wait_for_metadata_signal('/tempZone/home/rods', 'irods::signal_*signum');
+
+            # Log a message to let the admin know the metadata signal was received.
+            writeLine('serverLog', 'Received signal! (irods::signal_*signum)');
+
+            # Let the next delay rule know it's time to do actual work!
+            # We do that by attaching the correct metadata to the collection.
+            *next_signum = *signum + 1;
+            writeLine('serverLog', 'Signaling next delay rule (irods::signal_*next_signum) ...');
+            msiModAVUMetadata('-C', '/tempZone/home/rods', 'add', 'irods::signal_*next_signum', 'unused', '');
+        }
+    }
+}
+```
+
+### Let's see it in action!
+
+!!! Note
+    Remember, this example assumes you can modify metadata on `/tempZone/home/rods`. Feel free to change the collection used throughout the example to one that you can access and manipulate.
+
+If you haven't already, add the rule code above to `/etc/irods/core.re`.
+
+We need to run the example. To do that, run the following commands:
+```bash
+$ irule -r irods_rule_engine_plugin-irods_rule_language-instance 'launch_synchronized_delay_rules_example' null ruleExecOut
+$ imeta -C /tempZone/home/rods 'irods::signal_0' 'unused'
+```
+
+Now, inspect the log file at `/var/log/irods/irods.log`. You should see log messages similar to the ones below. Notice the value of **log_message**.
+```json
+{
+  "log_category": "legacy",
+  "log_facility": "local0",
+  "log_level": "info",
+  "log_message": "writeLine: inString = Received signal! (irods::signal_0)\n",
+  "request_api_name": "EXEC_RULE_EXPRESSION_AN",
+  "request_api_number": 1206,
+  "request_api_version": "d",
+  "request_client_user": "rods",
+  "request_host": "127.0.0.1",
+  "request_proxy_user": "rods",
+  "request_release_version": "rods4.3.0",
+  "server_host": "38832fb94fff",
+  "server_pid": 1919951,
+  "server_timestamp": "2022-08-23T12:17:50.161Z",
+  "server_type": "agent"
+}
+{
+  "log_category": "legacy",
+  "log_facility": "local0",
+  "log_level": "info",
+  "log_message": "writeLine: inString = Signaling next delay rule (irods::signal_1) ...\n",
+  "request_api_name": "EXEC_RULE_EXPRESSION_AN",
+  "request_api_number": 1206,
+  "request_api_version": "d",
+  "request_client_user": "rods",
+  "request_host": "127.0.0.1",
+  "request_proxy_user": "rods",
+  "request_release_version": "rods4.3.0",
+  "server_host": "38832fb94fff",
+  "server_pid": 1919951,
+  "server_timestamp": "2022-08-23T12:17:50.162Z",
+  "server_type": "agent"
+}
+{
+  "log_category": "legacy",
+  "log_facility": "local0",
+  "log_level": "info",
+  "log_message": "writeLine: inString = Received signal! (irods::signal_1)\n",
+  "request_api_name": "EXEC_RULE_EXPRESSION_AN",
+  "request_api_number": 1206,
+  "request_api_version": "d",
+  "request_client_user": "rods",
+  "request_host": "127.0.0.1",
+  "request_proxy_user": "rods",
+  "request_release_version": "rods4.3.0",
+  "server_host": "38832fb94fff",
+  "server_pid": 1919950,
+  "server_timestamp": "2022-08-23T12:17:51.158Z",
+  "server_type": "agent"
+}
+{
+  "log_category": "legacy",
+  "log_facility": "local0",
+  "log_level": "info",
+  "log_message": "writeLine: inString = Signaling next delay rule (irods::signal_2) ...\n",
+  "request_api_name": "EXEC_RULE_EXPRESSION_AN",
+  "request_api_number": 1206,
+  "request_api_version": "d",
+  "request_client_user": "rods",
+  "request_host": "127.0.0.1",
+  "request_proxy_user": "rods",
+  "request_release_version": "rods4.3.0",
+  "server_host": "38832fb94fff",
+  "server_pid": 1919950,
+  "server_timestamp": "2022-08-23T12:17:51.158Z",
+  "server_type": "agent"
+}
+{
+  "log_category": "legacy",
+  "log_facility": "local0",
+  "log_level": "info",
+  "log_message": "writeLine: inString = Received signal! (irods::signal_2)\n",
+  "request_api_name": "EXEC_RULE_EXPRESSION_AN",
+  "request_api_number": 1206,
+  "request_api_version": "d",
+  "request_client_user": "rods",
+  "request_host": "127.0.0.1",
+  "request_proxy_user": "rods",
+  "request_release_version": "rods4.3.0",
+  "server_host": "38832fb94fff",
+  "server_pid": 1920152,
+  "server_timestamp": "2022-08-23T12:17:51.402Z",
+  "server_type": "agent"
+}
+{
+  "log_category": "legacy",
+  "log_facility": "local0",
+  "log_level": "info",
+  "log_message": "writeLine: inString = Signaling next delay rule (irods::signal_3) ...\n",
+  "request_api_name": "EXEC_RULE_EXPRESSION_AN",
+  "request_api_number": 1206,
+  "request_api_version": "d",
+  "request_client_user": "rods",
+  "request_host": "127.0.0.1",
+  "request_proxy_user": "rods",
+  "request_release_version": "rods4.3.0",
+  "server_host": "38832fb94fff",
+  "server_pid": 1920152,
+  "server_timestamp": "2022-08-23T12:17:51.402Z",
+  "server_type": "agent"
+}
+{
+  "log_category": "legacy",
+  "log_facility": "local0",
+  "log_level": "info",
+  "log_message": "writeLine: inString = Received signal! (irods::signal_3)\n",
+  "request_api_name": "EXEC_RULE_EXPRESSION_AN",
+  "request_api_number": 1206,
+  "request_api_version": "d",
+  "request_client_user": "rods",
+  "request_host": "127.0.0.1",
+  "request_proxy_user": "rods",
+  "request_release_version": "rods4.3.0",
+  "server_host": "38832fb94fff",
+  "server_pid": 1919952,
+  "server_timestamp": "2022-08-23T12:17:52.160Z",
+  "server_type": "agent"
+}
+{
+  "log_category": "legacy",
+  "log_facility": "local0",
+  "log_level": "info",
+  "log_message": "writeLine: inString = Signaling next delay rule (irods::signal_4) ...\n",
+  "request_api_name": "EXEC_RULE_EXPRESSION_AN",
+  "request_api_number": 1206,
+  "request_api_version": "d",
+  "request_client_user": "rods",
+  "request_host": "127.0.0.1",
+  "request_proxy_user": "rods",
+  "request_release_version": "rods4.3.0",
+  "server_host": "38832fb94fff",
+  "server_pid": 1919952,
+  "server_timestamp": "2022-08-23T12:17:52.160Z",
+  "server_type": "agent"
+}
+{
+  "log_category": "legacy",
+  "log_facility": "local0",
+  "log_level": "info",
+  "log_message": "writeLine: inString = Received signal! (irods::signal_4)\n",
+  "request_api_name": "EXEC_RULE_EXPRESSION_AN",
+  "request_api_number": 1206,
+  "request_api_version": "d",
+  "request_client_user": "rods",
+  "request_host": "127.0.0.1",
+  "request_proxy_user": "rods",
+  "request_release_version": "rods4.3.0",
+  "server_host": "38832fb94fff",
+  "server_pid": 1919953,
+  "server_timestamp": "2022-08-23T12:17:53.168Z",
+  "server_type": "agent"
+}
+{
+  "log_category": "legacy",
+  "log_facility": "local0",
+  "log_level": "info",
+  "log_message": "writeLine: inString = Signaling next delay rule (irods::signal_5) ...\n",
+  "request_api_name": "EXEC_RULE_EXPRESSION_AN",
+  "request_api_number": 1206,
+  "request_api_version": "d",
+  "request_client_user": "rods",
+  "request_host": "127.0.0.1",
+  "request_proxy_user": "rods",
+  "request_release_version": "rods4.3.0",
+  "server_host": "38832fb94fff",
+  "server_pid": 1919953,
+  "server_timestamp": "2022-08-23T12:17:53.169Z",
+  "server_type": "agent"
+}
+```
+
+## Naming Schemes and Conventions 
+
+This section covers naming techniques and conventions that help in enforcing naming conventions.
+
+### Metadata Attribute Names
+
+Use prefixes to introduce different namespaces. Below is an example that generates metadata attribute names with a special prefix. The prefix itself isn't special, but it allows implementations to reuse variable names without creating collisions.
+
+```python
+FANCY_NAMESPACE = "fancy"; # The "fancy" namespace string.
+
+# A helper rule that returns the input string with a prefix of the fancy namespace string.
+make_fancy(*name)
+{
+    FANCY_NAMESPACE ++ "::*name";
+}
+
+generate_and_use_fancy_metadata_attribute_names
+{
+    # Generate metadata attribute names under the "fancy" namespace.
+    *attr_file_path = make_fancy("file_path"); # Creates "fancy::file_path".
+    *attr_file_size = make_fancy("file_size"); # Creates "fancy::file_size".
+
+    # Do something with the variables ...
+}
+```
+
+This can be used to generate keys for `temporaryStorage` as well. See [Sharing data across PEPs](#sharing-data-across-peps) for details about using `temporaryStorage`.
+
+### Rule Names
+
+When a group of rules are related, include a rule name prefix to indicate that. You can find several examples of this in other programming languages.
+
+Below is an example that demonstrates how this may look. The rules presented make up an imaginary Native Rule Language synchronization library. In this example, the rule name prefix, `mutex_` is used to provide signal to the person writing the policy that these rules are related. The body of the rules has been left out to not take away from the primary focus of the example.
+
+```python
+#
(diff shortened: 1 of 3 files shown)


docs/developers/library_examples.md
```diff
@@ -0,0 +1,613 @@
+# Library Examples
+
+## C
+
+Coming Soon ...
```
trel, line 5: still true?

docs/policy_cookbook/metadata_based_synchronization.md
```diff
@@ -0,0 +1,10 @@
+# Metadata-based Locking
```
trel, line 1: this title does not match the filename - so one or the other should probably get updated.

mkdocs.yml
```diff
@@ -69,9 +69,16 @@ nav:
     - Metadata: 'icommands/metadata.md'
     - Tickets: 'icommands/tickets.md'
     - Groupadmin: 'icommands/groupadmin.md'
+ - Policy Cookbook:
+    - Home: 'policy_cookbook/index.md'
```
trel, line 73: not sure about 'Home'... but don't have a better suggestion at the moment...

trel (commented):
does this render the left sidebar correctly now?

korydraughn:
> does this render the left sidebar correctly now?

Yes

docs/policy_cookbook/metadata_based_synchronization.md
korydraughn: Will correct that. The filename is the correct name.

Still a WIP

mkdocs.yml
korydraughn: Same here

docs/developers/library_examples.md
korydraughn: Maybe. I'm still thinking about this one.

Doxygen may be enough for the C APIs. However, it does seem like a good idea to provide examples around common operations. Some users will be afraid to use C++ due to complexity. For that reason, C may be a better fit. Here are a few things I feel would be good to have here for everyone, including the core dev team:
- Connecting to iRODS
- Iterating over a collection
- Using GenQuery

The Library Examples section can serve as a quick reference for various things instead of requiring devs to look through Doxygen documentation.

trel: reasonable list.   i think that makes sense.

mkdocs.yml
alanking: Here are some suggestions, which may or may not be better...
 - Index
 - Table of Contents
 - Overview
 - Introduction

Alternatively, this could all go under one page (much like the library examples or troubleshooting pages) and stuck under System Overview or even grow another section for administrators.

docs/policy_cookbook/index.md
```diff
@@ -0,0 +1,13 @@
+# The iRODS Policy Cookbook
+
+This section is all about presenting techniques to various policy-based situations encountered in the iRODS ecosystem.
+
+Overtime, we hope this cookbook will grow and become a valuable resource for new and existing users of iRODS.
```
alanking, line 5: Overtime -> "Over time"

docs/developers/library_examples.md
```diff
@@ -0,0 +1,614 @@
...
+
+## C++
+
+### client_connection
+
+Available Since: v4.2.9
```
alanking, line 11: Any advantage to linking to Doxygen pages from here? Is that even possible?

alanking (commented):
Very exciting

docs/developers/library_examples.md
korydraughn: Sounds like a good idea to me.

trel: yes, you can link to it, same as in the sidebar...    if you need something more targeted, then I'm not sure how stable the generated URLs are over time/versions.

docs/policy_cookbook/metadata_based_synchronization.md
korydraughn: Fixed title

docs/policy_cookbook/index.md
korydraughn: fixed

docs/policy_cookbook/metadata_based_synchronization.md
trel: filename itself?

korydraughn:
> I’m not familiar with the second parameter…

`msiSleep('seconds', 'microseconds')`

mkdocs.yml
korydraughn: Updated navigation.

Added section called `Administrators` and placed the `Policy Cookbook` under that.

`System Overview` could work, but the word, `Overview`, doesn't fit the information IMO.

Thoughts?

docs/developers/library_examples.md
```diff
@@ -0,0 +1,960 @@
...
+- replica_proxy
+- data_object_proxy
+- replica_state_table
+- replica_access_table
+- logical_locking
+- configuration_reload
+- at_scope_exit
+- scoped_client_identity
+- scoped_client_permission
+- scoped_priviledged_client
```
korydraughn, line 960: These can be individual issues.

korydraughn:
The information in this files is probably a good starting point.

I need to either remove the http links or make sure they stay up to date without us having to watch them constantly.

docs/developers/library_examples.md
```diff
@@ -0,0 +1,960 @@
...
+    auto conn = // Our iRODS connection.
+
+    // Instantiates a new transport object which uses the iRODS protocol to read and
+    // write bytes into a data object. Transport objects are designed to be used by IOStreams
+    // objects such as dstream. "default_transport" is a wrapper around the iRODS C API for
+    // reading and writing data objects.
+    //
+    // You can add support for more transport protocols by implementing the following interface:
+    //
+    //     https://github.com/irods/irods/blob/master/lib/core/include/transport/transport.hpp
```
trel, line 590: now 'main' branch?

docs/developers/library_examples.md
```diff
@@ -0,0 +1,960 @@
...
+    // This will also write data into the data object. This is slower than the previous method
+    // because stream operators format data.
+    out << "Here is some more data ...\n";
+}
+
+void read_from_data_object()
+{
+    namespace io = irods::experimental::io;
+
+    auto conn = // Our iRODS connection.
```
trel, line 619: nothing on right hand side?

docs/developers/library_examples.md
```diff
@@ -0,0 +1,960 @@
...
+Demonstrates how to construct query iterators via an `irods::experimental::query_builder`.
+
+```c++
+#include <irods/query_builder.hpp>
+
+#include <vector>
+
+void make_query()
+{
+    auto conn = // Our iRODS connection.
```
trel, line 728: same

docs/developers/library_examples.md
```diff
@@ -0,0 +1,960 @@
...
+Demonstrates how to use `rclOpenCollection`, `rclReadCollection` and `rclCloseCollection`.
+
+```cpp
+#include <irods/miscUtil.h>
+
+#include <iostream>
+
+void iterating_over_a_collection()
+{
+    RcComm* conn = // Our iRODS connection.
```
trel, line 73: one of 8 in this page - no right hand side?

docs/developers/library_examples.md
```diff
@@ -0,0 +1,960 @@
...
+    // multiplier. These options are completely optional. The last result will be
+    // returned to the call site. All intermediate results will be lost.
+    //
+    // The most important thing to understand about this function is the function-like
+    // object that will be invoked. It is up to the developer to instruct "with_durability"
+    // of when the set of operations have succeeded or failed, etc.
+    //
+    // See the following for more details:
+    //
+    //     https://github.com/irods/irods/blob/master/lib/core/include/with_durability.hpp
```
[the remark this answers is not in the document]
korydraughn: These URLs may be cut from the final version.

docs/developers/library_examples.md
korydraughn: That is correct. The examples are not meant to be runnable as they are. For some things, it is on the reader to fill in gaps.

The library examples also contain examples for connecting to an iRODS server so it shouldn't be difficult for the reader to add that.

docs/developers/library_examples.md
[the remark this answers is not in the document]
korydraughn: removed

docs/developers/library_examples.md
korydraughn: removed

docs/developers/library_examples.md
korydraughn: I'll add a link to the doxygen documentation at the top of the page. Like @trel said, it may be troublesome to manage links if they change between versions. We can also revisit this later.

docs/developers/library_examples.md
```diff
@@ -0,0 +1,971 @@
...
+    // value to the string "2". The string is copied into the object.
+    kvp[REPL_NUM_KW] = "2";
+
+    // This proxy type also supports checking if the kevValuePair_t contains a specific key.
+    if (kvp.contains(RESC_NAME_KW)) {
+        // Do something ...
+    }
+
+    // Extracting a value is easy too.
+    const std::string value = kvp[RESC_HIER_STR_KW];
```
alanking, line 703: As a best practice, maybe we should use `at` here? Or mention it?

docs/developers/library_examples.md
```diff
@@ -0,0 +1,971 @@
...
+
+```c++
+#include <irods/irods_query.hpp>
+
+#include <fmt/format.h>
+
+void print_all_resource_names(RcComm& _conn)
+{
+    // Print all resource names known to iRODS.
+    for (auto&& row : irods::query<RcComm>{&_conn, "select RESC_NAME"}) {
```
alanking, line 721: So smol

docs/developers/library_examples.md
alanking: Maybe we can have `conn` be passed in like in the `print_all_resource_names` example under `query`?

alanking (commented):
I think having everything on two pages is much better, personally. Overall this is amazing, my suggestions are nice-to-haves

Should we bring this out of draft?

docs/developers/library_examples.md
korydraughn: Very good idea. Will change.

docs/administrators/policy_cookbook.md
```diff
@@ -0,0 +1,251 @@
...
+
+If you have suggestions on how to improve the cookbook or you've developed solutions that you feel could help others, please reach out to us by creating an issue at [https://github.com/irods/irods_docs](https://github.com/irods/irods_docs).
+
+## Synchronizing Delay Rules using Metadata
+
+This example demonstrates how to synchronize actively running delay rules using metadata.
+
+### How to do it ...
+
+```python
```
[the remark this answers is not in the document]
korydraughn: That's correct.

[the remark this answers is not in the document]
korydraughn: It would be great if we could create a color scheme for the rule language.

docs/developers/library_examples.md
korydraughn: Absolutely.

docs/developers/library_examples.md
```diff
@@ -0,0 +1,971 @@
...
+
+    GenQueryInp input{}; // Curly braces are equivalent to using std::memset to clear the object.
+
+    // Fetch the maximum number of rows for a single page.
+    // This is specific to iRODS. The maximum page size has nothing to do with the database.
+    input.maxRows = MAX_SQL_ROWS;
+
+    // Here's where we specify the columns we want to fetch, along with various options.
+    addInxIval(&input.selectInp, COL_R_RESC_ID,   ORDER_BY_DESC /* Sort by resource id in descending order */);
+    addInxIval(&input.selectInp, COL_R_RESC_NAME, 0 /* No special options on this column */);
```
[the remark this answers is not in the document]
korydraughn: I was trying to emphasize that those comments are specific to the parameter. Do you feel that isn't conveyed well enough?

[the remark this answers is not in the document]
alanking: I see what you mean. I can't think of a *better* way, but an alternative could be to place the comment off to the side or above and reference the 3rd input in each instance, it just kind of looked like a mistake to me when I was reading through. Maybe it's clear to others as it is.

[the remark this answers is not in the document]
korydraughn: I'm looking at the rendered version and I agree. I'll change it.

docs/developers/library_examples.md
korydraughn: Done.

docs/developers/library_examples.md
korydraughn: Done.

docs/developers/library_examples.md
korydraughn: Done.

docs/administrators/policy_cookbook.md
```diff
@@ -0,0 +1,251 @@
...
+make_fancy(*name)
+{
+    FANCY_NAMESPACE ++ "::*name";
+}
+
+generate_and_use_fancy_metadata_attribute_names
+{
+    # Generate metadata attribute names under the "fancy" namespace.
+    *avu_file_path = make_fancy("file_path"); # Creates "fancy::file_path".
+    *avu_file_size = make_fancy("file_size"); # Creates "fancy::file_size".
```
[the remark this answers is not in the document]
korydraughn: Done.

docs/administrators/policy_cookbook.md
```diff
@@ -0,0 +1,251 @@
...
+
+The prototype implementation of the Logical Quotas Rule Engine Plugin uses this technique. You can find the implementation at [https://github.com/irods/irods_policy_examples/blob/main/irods_policy_logical_quotas.re](https://github.com/irods/irods_policy_examples/blob/main/irods_policy_logical_quotas.re)
+
+Additional documentation can be found at [Using temporaryStorage in the iRODS Rule Language](/system_overview/tips_and_tricks/#using-temporarystorage-in-the-irods-rule-language).
+
+## Simulating User Quotas
+
+This example demonstrates how to simulate user quotas using group quotas.
+
+As of iRODS 4.3.0, support for user quotas has been partially disabled. Please consult `iadmin suq` and the release notes for more information.
```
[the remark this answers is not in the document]
korydraughn: Done.

docs/administrators/policy_cookbook.md
```diff
@@ -0,0 +1,251 @@
...
+
+First, we have to instruct iRODS to enforce quotas. Open `/etc/irods/core.re` and change the argument passed to `msiSetRescQuotaPolicy()` from `"off"` to `"on"`. The line you modified should look very similar to the one below.
+```python
+acRescQuotaPolicy { msiSetRescQuotaPolicy("on"); }
+```
+We recommend applying this change to all servers in the local zone. That guarantees that all servers enforce the quotas. Keep in mind that this is only a recommendation. You should use a testing environment to verify behavior if you decide not to follow this recommendation.
+
+iRODS does not update the quota information following user interaction. To do that, you're going to need to periodically tell iRODS to update the quota information. One way to do that is by running `iadmin cu` periodically. How you do that is up to you. You can use **cron** or any other tool you find convenient to use. The important thing is that the command runs.
+
+With that out the way, we can focus on applying the actual quota.
```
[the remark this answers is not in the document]
korydraughn: Done.

docs/administrators/policy_cookbook.md
```diff
@@ -0,0 +1,251 @@
...
+
+### How to do it ...
+
+#### Step 1: Enable Quota Enforcement
+
+First, we have to instruct iRODS to enforce quotas. Open `/etc/irods/core.re` and change the argument passed to `msiSetRescQuotaPolicy()` from `"off"` to `"on"`. The line you modified should look very similar to the one below.
+```python
+acRescQuotaPolicy { msiSetRescQuotaPolicy("on"); }
+```
+We recommend applying this change to all servers in the local zone. That guarantees that all servers enforce the quotas. Keep in mind that this is only a recommendation. You should use a testing environment to verify behavior if you decide not to follow this recommendation.
```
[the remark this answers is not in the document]
korydraughn: Done.

docs/developers/library_examples.md
```diff
@@ -0,0 +1,971 @@
...
+#include <irods/getRodsEnv.h>
+#include <irods/rodsClient.h>
+#include <irods/rcConnect.h>
+
+void connecting_to_an_irods_server()
+{
+    rodsEnv env;
+
+    if (const int ec = getRodsEnv(&env); ec != 0) {
+        // Failed to load the local environment information (i.e. irods_environment.json).
```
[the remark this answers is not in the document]
korydraughn: Done.

docs/developers/library_examples.md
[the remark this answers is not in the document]
korydraughn: Done.

docs/developers/library_examples.md
korydraughn: Done.

docs/administrators/policy_cookbook.md
```diff
@@ -0,0 +1,251 @@
...
+
+As of iRODS 4.3.0, support for user quotas has been partially disabled. Please consult `iadmin suq` and the release notes for more information.
+
+### How to do it ...
+
+#### Step 1: Enable Quota Enforcement
+
+First, we have to instruct iRODS to enforce quotas. Open `/etc/irods/core.re` and change the argument passed to `msiSetRescQuotaPolicy()` from `"off"` to `"on"`. The line you modified should look very similar to the one below.
+```python
+acRescQuotaPolicy { msiSetRescQuotaPolicy("on"); }
```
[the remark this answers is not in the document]
korydraughn: I'm not sure what that blurb would say. Do you have words that you feel should be mentioned here?

docs/developers/library_examples.md
korydraughn: True. I've updated the text to link against things that are stable.

mkdocs.yml
korydraughn: @trel @alanking Any thoughts on the navigation?

You can see the current rendering by going to kdd-ws.euro...org:8080

docs/developers/library_examples.md
korydraughn: Done.

docs/developers/library_examples.md
```diff
@@ -0,0 +1,971 @@
...
+    // statements in SQL.
+    //
+    // "ShowCollAcls" expects the absolute path to a collection as shown below.
+    char input_arg[] = "/tempZone/home/rods";
+    input.args[0] = input_arg;
+
+    // IMPORTANT: Specific queries can accept up to ten arguments!
+
+    // Fetch the maximum number of rows for a single page.
+    // This is specific to iRODS. The maximum page size has nothing to do with the database.
```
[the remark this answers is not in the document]
korydraughn: Done.

docs/developers/library_examples.md
```diff
@@ -0,0 +1,971 @@
...
+#include <fmt/format.h>
+
+void fetch_resource_information_using_a_general_query()
+{
+    RcComm* conn = // Our iRODS connection.
+
+    GenQueryInp input{}; // Curly braces are equivalent to using std::memset to clear the object.
+
+    // Fetch the maximum number of rows for a single page.
+    // This is specific to iRODS. The maximum page size has nothing to do with the database.
```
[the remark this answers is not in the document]
korydraughn: Done.

docs/developers/library_examples.md
```diff
@@ -0,0 +1,971 @@
...
+                break;
+
+            default:
+                fmt::print("Unknown object type.\n");
+                break;
+        }
+    }
+
+    // The "rcl*" collection functions manage memory for us. These functions will free any heap
+    // allocated memory for us. Attempting to free any memory return from "rclReadCollection"
```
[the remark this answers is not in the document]
korydraughn: Done.

docs/developers/library_examples.md
```diff
@@ -0,0 +1,971 @@
...
+        rcDisconnect(conn);
+        return;
+    }
+
+    //
+    // Use the "conn" pointer to execute operations.
+    //
+
+    // When done, call rcDisconnect() so that the server knows the client is finished.
+    // This lets the server know it is okay to deallocate resources.
```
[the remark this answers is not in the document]
korydraughn: Done.

docs/administrators/policy_cookbook.md
```diff
@@ -0,0 +1,251 @@
...
+  "request_proxy_user": "rods",
+  "request_release_version": "rods4.3.0",
+  "server_host": "38832fb94fff",
+  "server_pid": 24234,
+  "server_timestamp": "2022-07-24T16:39:24.348Z",
+  "server_type": "agent"
+}
+```
+
+The prototype implementation of the Logical Quotas Rule Engine Plugin uses this technique. You can find the implementation at [https://github.com/irods/irods_policy_examples/blob/main/irods_policy_logical_quotas.re](https://github.com/irods/irods_policy_examples/blob/main/irods_policy_logical_quotas.re)
```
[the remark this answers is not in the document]
korydraughn: Done.

docs/administrators/policy_cookbook.md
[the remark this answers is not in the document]
trel: how's this...

> The original RescQuota system was geared to count bytes on storage resources... either per user or group for a single resource, or per user or group across all resources.  Since 4.3.0, user quotas have been removed (user and group quotas could not be enforced simultaneously - see irods/irods#4481)- but group quotas can still be set and enforced across all resources.  In this case, we're interested in setting a quota for a group which contains a single user (equivalent to setting a user quota).

Maybe we should start referring to this quota system as "Physical Quotas"...  wondering out loud if we can pull them out into a separate rule engine plugin...   gah, that sounds like a lot of work.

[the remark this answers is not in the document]
korydraughn: > The original RescQuota system was geared to count bytes on storage resources... either per user or group for a single resource, or per user or group across all resources. Since 4.3.0, user quotas have been removed (user and group quotas could not be enforced simultaneously - see irods/irods#4481)- but group quotas can still be set and enforced across all resources. In this case, we're interested in setting a quota for a group which contains a single user (equivalent to setting a user quota).

My gut reaction is that feels like something that should be linked to rather than placed directly in the cookbook. Although, technical books tend to have little pockets of text like that at various places.

I'm going to add it to see what it's like in the rendered version.

> Maybe we should start referring to this quota system as "Physical Quotas"... wondering out loud if we can pull them out into a separate rule engine plugin... gah, that sounds like a lot of work.

Why "Physical Quotas"? I feel we can externalize them into a REP. Unlike Logical Quotas, this system does not react to data movement in real time, so it may be simpler to implement. Removing the code from the server is probably where the challenge lies.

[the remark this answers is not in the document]
trel: That's fair about living somewhere else and linking... but, not sure where that would be yet.

--

Physical Quotas, rather than Logical Quotas.   Agreed that removing the `iadmin` subcommands would be a hairy process. And providing a(n automatic) transition/upgrade process is not clear either...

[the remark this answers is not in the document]
alanking: I mean, I just legitimately didn't know why we had to turn on resc quotas to enable this feature. :)

A single sentence like: "For historical reasons, this policy enables tracking quota usage." Or something like that would be fine by me. Fleshing that out more as @trel did could be valuable too, of course.

docs/administrators/policy_cookbook.md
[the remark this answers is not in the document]
korydraughn: Marking as resolved. We can always change the language used for code samples.

docs/administrators/policy_cookbook.md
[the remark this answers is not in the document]
korydraughn: Sure. I agree that @trel's statement needs to be added to the documentation. I may look into adding that somewhere via this PR.

Is this about protecting the reader from confusion due to the name of the microservice and NREP rule?

docs/administrators/policy_cookbook.md
```diff
@@ -0,0 +1,251 @@
...
+
+Let's say we have the following rules defined in `/etc/irods/core.re`.
+
+Pay close attention to the use of `temporaryStorage`. It is the bridge between PEPs within the Native Rule Engine Plugin.
+
+```python
+pep_api_data_obj_put_pre(*INSTANCE_NAME, *COMM, *DATA_OBJ_INPUT, *BYTES_BUFFER, *PORTAL_OPR_OUTPUT)
+{
+    # Capture the logical path of the PUT operation.
+    # We can use it later when writing messages to the log file.
```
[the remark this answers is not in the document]
korydraughn: Well, this info is available in the post PEP because of the `*DATA_OBJ_INPUT`.

I'll make sure to mention that the pre PEP always fires before the post PEP. I will also include a link to the dynamic PEPs page.

docs/administrators/policy_cookbook.md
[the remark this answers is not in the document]
trel: yes, the Resc in the name.... when we're attempting to set a 'user' quota.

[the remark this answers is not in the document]
korydraughn: I thought so. Take a look at the current write-up and let me know what you think. It is closer to what @alanking proposed.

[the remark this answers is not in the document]
alanking: I think it is sufficient, yep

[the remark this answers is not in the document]
trel: Agreed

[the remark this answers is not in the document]
korydraughn: Cool. Marking as resolved.

mkdocs.yml
trel: i like the navigation so far.   I think it's in the right place for now.

korydraughn: Excellent.

korydraughn:
I think this is in a good spot to be merged.

If there aren't any objections, then I'll go ahead and add the #'s.

trel:
agreed.

korydraughn:
Added #'s

korydraughn:
Don't merge yet. I need to update one thing.

korydraughn:
It's ready to be merged now.

Had to update the sync delay rules example.

alanking (approved):

PR #177, opened 2022-11-01
[irods/irods6557] Added information about detached mode

JustinKyleJames:


docs/plugins/composable_resources.md
```diff
@@ -276,6 +276,29 @@ acPostProcForDataCopyReceived(*leaf_resource) {
...
+
+```
+irods@hostname:~/ $ iadmin mkresc detached_resc3 unixfilesystem hostname.example.org:/common/mount/point "host_mode=detached;host_list=host2.example.org,host3.example.org"
+```
+
+The host assigned to the host field for the resource is assumed to be able to serve up any request, so in the above case either hostname, host2, or host3 may serve up the request.
+
+The following are some rules around using detached mode:
+
+1. If host_list does not exist but host_mode=detached, all hosts are assumed to be able to service requests for the resource.
```
korydraughn, line 297: `but host_mode=detached,` -> `but host_mode=detached does,`?

Is this statement true if host_list is empty (i.e. `host_mode=detached;host_list=`)?

docs/plugins/composable_resources.md
```diff
@@ -276,6 +276,29 @@ acPostProcForDataCopyReceived(*leaf_resource) {
...
+2. If host_list does exist, only the hosts in either this list or the host in the resource host field will be able to service requests for the resource.
+3. If a host list exists and the current host is not in the list, redirects will go to the host listed in the resource host field as usual. If that host is down the request will fail even if other hosts in host_list are up.
+4. The path to the resource vault path must be identical on every host that services requests for the resource. This vault path must be in a mount to a common filesystem.
+5. Registrations outside the vault are not allowed as we can't enforce that the path being registered is shared by all hosts that service requests for the resource.
```
korydraughn, line 301: Consider splitting this statement into multiple statements.

korydraughn (commented):
Very nice. Just a few things to consider.

docs/plugins/composable_resources.md
```diff
@@ -276,6 +276,29 @@ acPostProcForDataCopyReceived(*leaf_resource) {
 }
 ```
 
+Detached mode was added as a configuration option for UnixFileSystem resources in release 4.3.1.  When detached mode is enabled, any server may serve up requests for the resource as long as all servers share a common mounted vault.
+
+To enable detached mode, the host_mode setting in the context string should be set to "detached" as in the following example:
+
+```
+irods@hostname:~/ $ iadmin mkresc detached_resc3 unixfilesystem hostname.example.org:/common/mount/point "host_mode=detached"
```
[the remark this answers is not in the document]
JustinKyleJames: That was a typo.  It has been removed.

docs/plugins/composable_resources.md
```diff
@@ -276,6 +276,29 @@ acPostProcForDataCopyReceived(*leaf_resource) {
...
+```
+
+If an administrator wants to restrict the number of hosts that may serve the request, a host_list parameter with a comma separated list of hosts may be added to the context string.
```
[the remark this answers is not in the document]
JustinKyleJames: Done

docs/plugins/composable_resources.md
```diff
@@ -276,6 +276,29 @@ acPostProcForDataCopyReceived(*leaf_resource) {
...
+irods@hostname:~/ $ iadmin mkresc detached_resc3 unixfilesystem hostname.example.org:/common/mount/point "host_mode=detached;host_list=host2.example.org,host3.example.org"
+```
+
+The host assigned to the host field for the resource is assumed to be able to serve up any request, so in the above case either hostname, host2, or host3 may serve up the request.
+
+The following are some rules around using detached mode:
+
+1. If host_list does not exist but host_mode=detached, all hosts are assumed to be able to service requests for the resource.
+2. If host_list does exist, only the hosts in either this list or the host in the resource host field will be able to service requests for the resource.
+3. If a host list exists and the current host is not in the list, redirects will go to the host listed in the resource host field as usual. If that host is down the request will fail even if other hosts in host_list are up.
```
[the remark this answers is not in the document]
JustinKyleJames: Done

docs/plugins/composable_resources.md
JustinKyleJames: It will behave as if detached mode is disabled.  The host in the resource configuration will essentially be the only host in the host list.

docs/plugins/composable_resources.md
JustinKyleJames: Done

docs/plugins/composable_resources.md
```diff
@@ -276,6 +271,31 @@ acPostProcForDataCopyReceived(*leaf_resource) {
 }
 ```
 
+#### Detached Mode Configuration
+
+Detached mode was added as a configuration option for UnixFileSystem resources in release 4.3.1.  When detached mode is enabled, any server may serve up requests for the resource as long as all servers share a common mounted vault.
+
+To enable detached mode, the **host_mode** setting in the context string should be set to **detached** as in the following example:
```
alanking, line 278: What happens if we set it to "attached" or "banana" or empty string? Worth noting?

docs/plugins/composable_resources.md
```diff
@@ -276,6 +271,31 @@ acPostProcForDataCopyReceived(*leaf_resource) {
...
+irods@hostname:~/ $ iadmin mkresc detached_resc unixfilesystem hostname.example.org:/common/mount/point "host_mode=detached;host_list=host2.example.org,host3.example.org"
+```
+
+The host assigned to the host field for the resource is assumed to be able to serve up any request, so in the above case either hostname, host2, or host3 may serve up the request.
+
+The following are some rules around using detached mode:
+
+1. If host_list does not exist, all hosts are assumed to be able to service requests for the resource.
+2. If host_list does exist, only the hosts in either this list or the host in the resource host field will be able to service requests for the resource.
+3. If a host_list exists and the current host is not in the list, redirects will go to the host listed in the resource host field as usual. If that host is down, the request will fail even if other hosts in host_list are up.
```
alanking, line 296: > If a host_list exists and the current host is not in the list, redirects will go to the host listed in the resource host field as usual.

Any thoughts on this alternative wording?

> If host_list exists and the client-connected server is not in host_list, the request will be redirected to the host listed in the resource host field as is done in attached mode.

docs/plugins/composable_resources.md
```diff
@@ -276,6 +271,31 @@ acPostProcForDataCopyReceived(*leaf_resource) {
...
+```
+irods@hostname:~/ $ iadmin mkresc detached_resc unixfilesystem hostname.example.org:/common/mount/point "host_mode=detached;host_list=host2.example.org,host3.example.org"
+```
+
+The host assigned to the host field for the resource is assumed to be able to serve up any request, so in the above case either hostname, host2, or host3 may serve up the request.
+
+The following are some rules around using detached mode:
+
+1. If host_list does not exist, all hosts are assumed to be able to service requests for the resource.
+2. If host_list does exist, only the hosts in either this list or the host in the resource host field will be able to service requests for the resource.
```
alanking, line 295: does exist -> exists

alanking (commented):
A fine write-up

docs/plugins/composable_resources.md
JustinKyleJames: Anything other than "detached" will default to attach mode.

docs/plugins/composable_resources.md
JustinKyleJames: I like it.  Will change.

docs/plugins/composable_resources.md
alanking: Okay. I feel that this may be worth a mention in the actual docs.

trel: agreed, just a sentence that says...

If "host_mode" is not set to "detached", the unixfilesystem plugin will default to "attached" mode.

JustinKyleJames: I added the sentence in.  Please take a look at where I added it and the use of quotes.  (I didn't include quotes around "host_mode" as we didn't do it elsewhere nor highlighted it because it has already been highlighted.  I also didn't use quotes around "attached" because we haven't stated that it is a setting with meaning.)

trel:
good enough to squash for me.

docs/plugins/composable_resources.md
```diff
@@ -276,6 +271,33 @@ acPostProcForDataCopyReceived(*leaf_resource) {
...
+
+Detached mode was added as a configuration option for UnixFileSystem resources in release 4.3.1.  When detached mode is enabled, any server may serve up requests for the resource as long as all servers share a common mounted vault.
+
+To enable detached mode, the **host_mode** setting in the context string should be set to **detached** as in the following example:
+
+```
+irods@hostname:~/ $ iadmin mkresc detached_resc unixfilesystem hostname.example.org:/common/mount/point "host_mode=detached"
+```
+
+If host_mode is not set to "detached", the unixfilesystem plugin will default to attached mode.
```
korydraughn, line 284: Probably should remove the double quotes since they aren't used in other places for this documentation.

JustinKyleJames: It was my thinking that the context is a bit different here.  I wanted to emphasize that it should be exactly this.  That said the check is not case sensitive.

JustinKyleJames: I went ahead and bolded host_mode and detached to emphasize these are exact settings.  I also add "(not case sensitive)" after detached.

korydraughn:
I think this is ready. Let's squash it without the pound.

JustinKyleJames:
Done

alanking (approved):
Let's # it

PR #181, opened 2022-11-04
[180] add read/write weights of zero to passthru

trel:


docs/plugins/composable_resources.md
```diff
@@ -197,6 +197,13 @@ To modify an existing passthru resource to be written to only after other eligib
 irods@hostname:~/ $ iadmin modresc newResc context 'write=0.4;read=1.0'
 ```
 
+Setting a `read` or `write` weight to `0` will prevent that action from occurring on the branch of the resource tree under the passthru (because a `0` vote signals that it cannot service the incoming request).  The following will 'turn off' the resource `newResc` (a policy-based way of marking a resource "down"):
```
alanking, line 200: Should this say 0.0 instead of 0? I'm guessing 0 also works, but the example below says 0.0.

Also, would it be worth mentioning that this "turns off" the entire subtree under `newResc` (not just `newResc`)?

docs/plugins/composable_resources.md
```diff
@@ -197,6 +197,13 @@ To modify an existing passthru resource to be written to only after other eligib
...
+
+
+```
+irods@hostname:~/ $ iadmin modresc newResc context 'write=0.0;read=0.0'
```
alanking, line 204: Should we show what this looks like, or would that possibly change over time? Something like...
```bash
irods@hostname:~/ $ ilsresc
<tree>
irods@hostname:~/ $ iadmin modresc newResc context 'write=0.0;read=0.0'
irods@hostname:~/ $ iput -R newResc newfile
<error>
irods@hostname:~/ $ iget -R newResc oldfile_with_one_replica
<error>
```

docs/plugins/composable_resources.md
```diff
@@ -197,6 +197,13 @@ To modify an existing passthru resource to be written to only after other eligib
 irods@hostname:~/ $ iadmin modresc newResc context 'write=0.4;read=1.0'
 ```
 
+Setting a `read` or `write` weight to `0` will prevent that action from occurring on the branch of the resource tree under the passthru (because a `0` vote signals that it cannot service the incoming request).  The following will 'turn off' the resource `newResc` (a policy-based way of marking a resource "down"):
+
+
```
korydraughn, line 202: These lines can be removed.

docs/plugins/composable_resources.md
korydraughn, line 200: Could replace the single quotes with `**` so that the text is bold?
Or replace the single quotes with `_` so that the text is italic?

docs/plugins/composable_resources.md
korydraughn: This seems like something that could be good for the policy cookbook.

trel: definitely - showing results is ... out of scope for this part of the docs i think

docs/plugins/composable_resources.md
korydraughn: Consistency is good. I vote for 0.0 since the example uses that.

trel: yeah, went back and forth over it - but looking at it again... seems like a typo rather than intentional.

will use 0.0... a bit more refined :)

will mention "(and subtree)"

trel: could be - it was all consistent at the 4.2.0 release... and then, things got 'more'.

i'll stare it again.   i kind of like the 'code' treatment of the backticks...

korydraughn: No no, I'm only referring to the `'turn of'` phrase.

Everything else is using backticks.

trel: ohhhh, yes - I'll try double quotes instead of single, first.

docs/plugins/composable_resources.md
alanking: Makes sense to me.

korydraughn:
Does setting these weights to 0.0 block ALL requests on the resource or a specific set of requests?

trel:
well, no, I think unlink and register and unregister and … might still get through.   Rename?

All these would be addressed by adding more verbs to the context string parser (and getting rid of marking things ‘down’… https://github.com/irods/irods/issues/4508)

docs/plugins/composable_resources.md
trel: tried zero rather than 0.0 in the running text.

trel: doubles seems good

docs/plugins/composable_resources.md
trel: removed one - kept one empty line before codeblock.

docs/plugins/composable_resources.md
```diff
@@ -197,6 +197,12 @@ To modify an existing passthru resource to be written to only after other eligib
 irods@hostname:~/ $ iadmin modresc newResc context 'write=0.4;read=1.0'
 ```
 
+Setting a 'read' or 'write' weight to zero will prevent that action from occurring on the branch of the resource tree under the passthru (because a zero vote signals that it cannot service the incoming request).  The following will "turn off" the resource `newResc` (and its sub-tree), a policy-based way of marking a resource "down":
```
korydraughn, line 200: Why single quotes for `read` and `write`, but double quotes for `turn off` and `down`?

Also, until we have options for unlink, rename, checksum, etc, should we say this is `a policy-based way of marking a resource "down"`?

trel: read and write have single quotes above in the section prior.   and turn off and down are in double quotes in other places on the same page.

regarding the phrasing of "down"... no, I will remove it - it gets ahead of ourselves.

trel: went with consistent double quotes, and removed the policy-based phrase.

korydraughn: Cool cool

korydraughn (approved):
Looks good to me.

trel:
squashed and pushed.

PR #191, opened 2023-02-22
Documented missing specific query, DataObjInCollReCur, for MySQL (4-2-stable)

korydraughn:


trel:
seems good

korydraughn:
Going to test and make sure that copying the command works as expected. If it does, I'll add the pound.

korydraughn:
I've verified that the command in the documentation works.

I had to modify the testing environment so that it would stand up a 4.2.11 server backed by MySQL 8 (basically had to change the collation and charset settings - PR will come soon).

Adding pound.

korydraughn:
#'d

korydraughn:
This may not be doable for 4.2.

As of right now, 4.2 setup fails against MySQL 8 unless the collation and charset are changed to something that doesn't cause the **row size too large** error. Otherwise, the admin must modify the database schema in a way that isn't supported by the 4.2 server.

trel (approved):
this seems correct

korydraughn:
Good. Will squash and pound shortly.

korydraughn:
Added pounds.

Issue #216, opened 2023-10-02
GitHub Actions builder image base (Ubuntu 18.04) is unsupported


alanking:
The recommendation is to update to Ubuntu 22.04, I suppose assuming that we stick with Ubuntu.

See https://github.com/actions/runner-images/issues/6002

PR #217, opened 2023-10-02
[#216] Update builder image

alanking:
Addresses #216 

We can't update to Ubuntu 22.04 until 4.3.1 is released. In the meantime, this seems to work both in GitHub Actions and on my local machine (docker version 24.0.6).


korydraughn (approved):
Looks fine to me.

Pound it.

PR #239, opened 2024-01-12
[irods/irods#4904] irodsPamAuthCheck: Add more docs (main)

alanking:
Addresses irods/irods#4904

My understanding is that we have no plans to update irodsPamAuthCheck itself. If anything, we would write a different tool for use in the future if we wanted to make any changes.

Also, happy to squash down to one if we're into it.


korydraughn (approved):
Very nice. Pound when ready.

korydraughn:
Your second commit message about the heading needs a tweak.

`so this commit move the` -> `so this commit moves the`

alanking:
Good catch. I've amended the commit message and squashed to one.

korydraughn (approved):
Ship it.

Issue #231, opened 2023-10-21
Broken links in Policy Cookbook page


korydraughn:
Links which reference other pages of docs.irods.org are broken.

The links are missing the version number (e.g. 4.3.1).

Issue #230, opened 2023-10-21
Broken link on Library Examples page


korydraughn:
The link to **Doxygen documentation** at the top of the page is broken.

PR #246, opened 2024-03-15
[#230,#231] Fix broken links

alanking:
Addresses #230 
Addresses #231

I did not transfer these issues to the 4.3.2 milestone in case we wanted to fix this in 4.3.1 documentation as well. Even if we don't, I wouldn't consider this a blocker for 4.3.2, so that is why I left them. Please let me know if we'd rather these issues be in the 4.3.2 milestone and I will transfer them and update the issue numbers appropriately.

Verified locally, but the problem does not manifest on locally running irods_docs sites, so it's hard to tell. I just copied how other places linked to these same pages in other places since those links work.


trel:
We can manually tweak the published/rendered 4.3.1 docs on the live site and just get these into the 4.3.2 milestone.

Seems easy enough.

alanking:
Okay, so... we do want to move these two issues into irods/irods and put them in the 4.3.2 milestone?

korydraughn (approved):
Looks good.

Pound it if you're happy with it.

trel:
Either is fine.  If these are in here soon/now, they'll make it into the 4.3.2 timeline regardless.

alanking:
Okay, I'm just going to leave these in this repository, then. Thanks

PR #259, opened 2024-05-22
[irods/irods#7382] Fix installation instructions for Postgres 15+

SwooshyCueb:
In service of irods/irods#7382


alanking (commented):
Seems good to me. Do these instructions work for older versions of Postgres? No is okay, but perhaps we should mention the minimum version somewhere... Or even put the old instructions in a separate section?

SwooshyCueb:
Yes, this works for older versions as well

alanking (approved):
Nice. # it.

SwooshyCueb:
\#'d

PR #269, opened 2024-07-09
[irods/irods7268] Clarify documentation on NREP condition behavior

FifthPotato:
First shot at this. Taking comments on whether it's detailed enough or not, etc.
Seems to work ok when tested locally via docker.


trel:
That seems better.    I think we run with it.

korydraughn (approved):
Much better than before.

Pound it.

alanking (approved):

FifthPotato:
Octothorpe'd.

PR #270, opened 2024-07-10
[irods/irods7268] Add extra warning to advise users against using on()

FifthPotato:
More words to explain why `on()` might do weird things; recommending `if()` as well.

### DIFF
--- a/docs/plugins/irods_rule_language.md
+++ b/docs/plugins/irods_rule_language.md
@@ -586,6 +586,9 @@ This Is Not A Valid Function Name or Rule Name
 
 ### Rule Condition
 
+!!! Note
+    Use of `on()` may lead to unexpected behavior. If all `on()` conditions within a rule fail, the rule will return an error. Consider use of `if()` for more familiar flow control.
+
 Rule conditions should be expressions of type `boolean`. The rule is executed only when the rule condition evaluates to true. There exist three failure conditions that will cause a rule failure error:
 
   1. All rule conditions evaluate to false.


docs/plugins/irods_rule_language.md
```diff
@@ -550,6 +550,8 @@ The syntax of a rule with a nontrivial rule condition is as follows:
 }
 ~~~
 
+&#x26A0;&#xFE0F; **Use of `on()` may lead to unexpected behavior!** If all rule conditions for a rule fail, the rule will return an error! Use `if()` for simple conditions.
```
trel, line 553: Consider the `!!! Note` block notation for this pullout.

example...
https://github.com/irods/irods_docs/blob/95425578aa9bcdce629096f6aacd42387e957604/docs/release_notes.md?plain=1#L2061-L2062

```suggestion
 !!! Note
     Use of `on()` may lead to unexpected behavior. If all `on()` conditions within a rule fail, the rule will return an error. Consider use of `if()` for more familiar flow control.
```

Also consider whether this would be better placed at the beginning of the ```### Rule Condition``` section a little farther down - near line 590?

alanking: Agreed. Even though `on()` is used here (and elsewhere, I'd imagine), it seems most appropriate to steer toward using `if()` in the conditional section.

docs/plugins/irods_rule_language.md
```diff
@@ -586,6 +586,9 @@ This Is Not A Valid Function Name or Rule Name
 
 ### Rule Condition
 
+!!! Note
+    **Use of `on()` may lead to unexpected behavior!** If all rule conditions for a rule fail, the rule will return an error! Use `if()` for simple conditions.
```
alanking, line 590: The rich diff looks a little strange here. "!!! Note" is actually being printed in addition to getting the indentation with the little bar on the left. Does this look right when it's rendered in the docs?

FifthPotato: Rich diff?
Here's what I see:
temp

korydraughn, line 590: Something about `for simple conditions.` doesn't sit right with me.

That phrase implies there's a way to exactly identify a simple condition vs a complex condition when in fact that is subjective.

What if we just said - "Prefer using `if()` instead of `on()`."

Thoughts?

alanking: Oh, I think the diff in github must render differently. What you showed looks better! Resolving this

trel: i had changed the wording up above in my suggestion.   search for 'familiar'...

korydraughn: The wording you provided is good too. I'm fine with using that.

FifthPotato: Oops, missed the additional wording change above. Added!

docs/plugins/irods_rule_language.md
```diff
@@ -586,6 +586,9 @@ This Is Not A Valid Function Name or Rule Name
 
 ### Rule Condition
 
+!!! Note
+    **Use of `on()` may lead to unexpected behavior!** If all rule conditions for a rule fail, the rule will return an error! Consider use of `if()` for more familiar flow control.
```
trel, line 590:
```suggestion
    Use of `on()` may lead to unexpected behavior. If all `on()` conditions within a rule fail, the rule will return an error. Consider use of `if()` for more familiar flow control.
```

korydraughn (commented):
Looks good to me.

If others agree, squash it.

trel:
squash em up.

FifthPotato:
Squished.

trel:
does not look like what you had before...  something has gone sideways during the squishing.

FifthPotato:
Uh oh, I did the git wrong. One more time?

trel:
i think we did it.

korydraughn (approved):
Pound it.

FifthPotato:
Octothorpe'd

korydraughn:
Merging.

Issue #291, opened 2024-11-05
Document iRODS authentication TTL default behavior


alanking:
The default behavior for TTL of iRODS "limited passwords" is to use the configured `password_min_time` in `R_GRID_CONFIGURATION`. This is not documented explicitly.

PR #302, opened 2024-12-20
[#291] Troubleshooting: Authentication TTL default behavior

alanking:
Addresses #291

### DIFF
--- a/docs/system_overview/troubleshooting.md
+++ b/docs/system_overview/troubleshooting.md
@@ -769,3 +769,16 @@ The environment of the service account is the means by which the server communic
 If the server *has not been* restarted after running `iexit`, `iinit` can be run with the service account `rodsadmin` password, and the service account's iRODS user can authenticate again and things return to normal.
 
 If the server *has been* restarted after running `iexit`, the server will stand up, but new connections cannot be established with it. Regardless, the service account can run `iinit` with the service account `rodsadmin` password. The `.irodsA` file will be generated file again after the connection to the server fails (may take a bit to timeout). The server can then be started again and things will return to normal.
+
+## Users are forced to re-authenticate after a few minutes
+
+If your users are authenticating via PAM (e.g. `pam_password` scheme) and find that they are being made to re-authenticate after only a few minutes, this section should provide an explanation and a way to remedy the situation.
+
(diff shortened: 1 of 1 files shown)


docs/system_overview/troubleshooting.md
```diff
@@ -769,3 +769,16 @@ The environment of the service account is the means by which the server communic
...
 
 If the server *has been* restarted after running `iexit`, the server will stand up, but new connections cannot be established with it. Regardless, the service account can run `iinit` with the service account `rodsadmin` password. The `.irodsA` file will be generated file again after the connection to the server fails (may take a bit to timeout). The server can then be started again and things will return to normal.
+
+## Users are forced to re-authenticate after a few minutes
+
+If your users are authenticating via PAM (e.g. `pam_password` scheme) and find that they are being made to re-authenticate after only a few minutes, this section should provide an explanation and a way to remedy the situation.
+
+An authenticated "session" for an iRODS user is managed through a Time-to-Live (TTL) parameter used by the authentication plugins. A session is said to "expire" after it has been valid for a specified TTL. For PAM authentication, sessions expire after the zone's configured `password_min_time` (in `R_GRID_CONFIGURATION` table) by default. The default `password_min_time` is 121 seconds. This explains the behavior described above.
```
trel, line 777:
```suggestion
An authenticated "session" for an iRODS user is managed through a Time-to-Live (TTL) parameter used by the authentication plugins. A session is said to "expire" after it has been valid for a specified TTL. For PAM authentication, the default behavior is for sessions to expire after the zone's configured `password_min_time` (in `R_GRID_CONFIGURATION` table) with a default `password_min_time` of 121 seconds. This explains the behavior described above.
```

trel: is that better?   hoping it is clearer... but... also, want to make sure it's still true.

alanking: In this case, I feel that it is slightly less clear...

It could be read to mean that `password_min_time` is separately configurable for the default behavior of PAM TTL. Or perhaps that the default behavior of expiring after the zone's configured `password_min_time` is connected to a particular value of `password_min_time`. See what I mean? I've bolded the part that makes it seem this way:

> the default behavior is for sessions to expire after the zone's configured `password_min_time` **with a default** `password_min_time` of 121 seconds

To me, it is clearer to keep the statements as separate sentences to make the disjoint nature of the default behaviors clear. The default behavior of PAM using a TTL of `password_min_time` is true regardless of the value (or the default value) of `password_min_time`.

Hope that makes sense. Maybe it's just me that read it that way.

trel: that's fine.  i struggled with the same thing.   carry on.  resolving.

alanking:
\#'d. Mergin

PR #311, opened 2025-02-20
[irods/irods#7773] Policy Cookbook: Access info for opened replicas

alanking:
Addresses irods/irods#7773


docs/administrators/policy_cookbook.md
```diff
@@ -798,3 +798,152 @@ The implementation can be modified to apply only to specific resources, collecti
...
+  },
+  "stage_flag": 0
+}
+```
+
+In order to access this information via the microservice, we need a handle to a JSON structure and a JSON pointer indicating which key's value we want. For example, if we wanted to know the creation time of the replica, we would use `"/data_object_info/data_create"`.
+
+In this example, we will implement a dynamic PEP for the DataObjRead API, which uses an `OpenedDataObjInp` structure as part of its signature. In this PEP, we do not know the logical path of the data object which is being read and the `OpenedDataObjInp` structure does not have this information. However, it does give us access to an L1 file descriptor. Using this, we can access information about the opened replica which already resides in the iRODS agent's memory. Here is the implementation:
+```python
+pep_api_data_obj_read_pre(*instance_name, *comm, *opened_data_obj_inp, *data_obj_read_out_bbuf) {
```
korydraughn, line 912: Move the opening curly brace to its own line so it matches the rest of the cookbook examples.

korydraughn (commented):
Very nice. Wording is good.

Found one thing. Once that's resolved, pound it.

trel:
very nice

alanking:
#'d, but will hold on merging juuuust in case

PR #316, opened 2025-03-12
[irods/irods#5952] Revert deprecation of Static Policy Enforcement Points

alanking:
Addresses irods/irods#5952


alanking:
#'d, mergin

PR #317, opened 2025-03-12
Updates for config documentation and PAM TTL handling (main)

korydraughn:


docs/system_overview/configuration.md
```diff
@@ -913,11 +913,15 @@ The maximum TTL of a randomly generated password in seconds. If a user attempts
 
 The minimum TTL of a randomly generated password in seconds. If a user attempts to authenticate with a TTL value that is less than `password_min_time`, the TTL is determined to be invalid and an error is returned. If `password_min_time` is configured to a value greater than `password_max_time` no passed-in TTL value will satisfy the system. Accepted values: [0..18446744073709552000). If a value outside of the acceptable range is used, a warning message will be logged for the administrator and the default value will be used. The default value is 121.
 
+!!! Note
+    If users are being forced to re-authenticate via PAM frequently, you may need to adjust this option. See [Users are forced to re-authenticate after a few minutes](../../system_overview/troubleshooting/#users-are-forced-to-re-authenticate-after-a-few-minutes) for more information.
```
alanking, line 917: Maybe mention `CAT_PASSWORD_EXPIRED` here somewhere so that it can be found in a search. Might be worth including that error message explicitly in the Troubleshooting section as well.

alanking (commented):
Seems good.

docs/system_overview/configuration.md
korydraughn: done

korydraughn:
Issue for `host` property has been created. Will attach that once PR is approved.

docs/system_overview/configuration.md
```diff
@@ -199,6 +193,10 @@ This file defines the behavior of the server Agent that answers individual reque
         // ... Additional Entries ...
     ],
 
+    // The fully qualified domain name or hostname which the server identifies as.
```
trel, line 196: can we rephrase to not end with a preposition?

not convinced this is the answer....
```
The fully qualified domain name or hostname which identifies the server 
```

korydraughn: Yeah, I don't know that the suggestion indicates what the property means. The property defines what the server would return if `hostname` were a command provided by the iRODS API. Another way to look at it is - all invocations of `gethostname()` in the server code should be replaced by the value of this property.

Does that help in explaining what I'm attempting to capture?

trel: yes, i understand what it's supposed to be.  just that the grammar was weird/awkward.

korydraughn: Hmm, maybe the suggestion IS what I'm looking for? The `host` property just needs to resolve to the server since one of its uses is to allow the server to connect to itself.

korydraughn: From the server code:

> // The host property in server_config.json defines the true identity of the local server.
// We cannot use localhost or the loopback address because the computer may have multiple
// network interfaces and/or hostnames which map to different IPs.

korydraughn: How about this?
```
The FQDN or hostname which represents the true identity of the server.
```

korydraughn: On second thought, your suggestion is good. The important thing is to capture the rules around it. That is - no localhost, must resolve to the local machine for reasons.

Will update.

korydraughn: Done.

docs/system_overview/troubleshooting.md
```diff
@@ -780,6 +780,9 @@ If the server *has been* restarted after running `iexit`, the server will stand
 
 If your users are authenticating via PAM (e.g. `pam_password` scheme) and find that they are being made to re-authenticate after only a few minutes, this section should provide an explanation and a way to remedy the situation.
 
+!!! Note
+    This situation is normally accompanied by a high occurrence of `CAT_PASSWORD_EXPIRED` in the server log.
```
alanking, line 784: Not sure if this deserves a `!!! Note` box, but I don't feel too strongly about that.

docs/system_overview/configuration.md
```diff
@@ -193,8 +193,12 @@ This file defines the behavior of the server Agent that answers individual reque
         // ... Additional Entries ...
     ],
 
-    // The fully qualified domain name or hostname which the server identifies as.
-    // The name defined by this setting must be resolvable by the server.
+    // The FQDN or hostname which identifies the server.
+    //
+    // The following rules apply:
+    // - Must not be localhost
```
alanking, line 199: Would it be clearer with quotation marks?
```suggestion
    // - Must not be the string "localhost"
```

alanking (commented):
Couple of other suggestions from me. Totally optional

docs/system_overview/configuration.md
trel: Agree

korydraughn: Will update.

docs/system_overview/troubleshooting.md
korydraughn: That's fair. Will adjust.

docs/system_overview/configuration.md
korydraughn: Done.

docs/system_overview/troubleshooting.md
korydraughn: Done.

korydraughn:
https://github.com/irods/irods/issues/7344 mentions CAT_INVALID_AUTHENTICATION and CAT_INVALID_USER. I'm going to add those as a note to the PAM section.

alanking (approved):
Looks good to me. Squash and # it

korydraughn:
Squashed/Pounded.

PR #321, opened 2025-03-13
[irods/irods#7957] Pluggable Rule Engine: Add Metadata Guard

alanking:
In service of irods/irods#7957

I just copied the README from the GitHub repository and added a blurb at the beginning... Not sure if we wanted something more (or less).


docs/plugins/pluggable_rule_engine.md
```diff
@@ -11,7 +11,6 @@ The iRODS Consortium currently supports the following rule engine plugins:
  - [C++ Default Policy Rule Engine Plugin](https://github.com/irods/irods/blob/main/plugins/rule_engines/src/cpp_default_policy.cpp)
  - [C++ Audit (AMQP) Rule Engine Plugin](https://github.com/irods/irods_rule_engine_plugin_audit_amqp)
  - [C++ Logical Quotas Rule Engine Plugin](https://github.com/irods/irods_rule_engine_plugin_logical_quotas)
- - [C++ Metadata Guard Rule Engine Plugin](https://github.com/irods/irods_rule_engine_plugin_metadata_guard)
```
korydraughn, line 14: Could make this point to the new section.

That would allow it to be listed with the other plugins.

korydraughn (commented):
Looks good.

docs/plugins/pluggable_rule_engine.md
trel: can we make it the same as the C++ default policy plugin three above?

logical quotas will have the same treatment at some point ...

docs/plugins/pluggable_rule_engine.md
```diff
@@ -293,3 +292,68 @@ The best practice for using both `delay()` and `remote()` [depends on the use ca
...
+            // - "user"
+            // - "group"
+            "type": "group",
+
+            // The name of the iRODS entity.
+            // For remote users, you must include the zone (e.g. "rods#tempZone").
+            "name": "rodsadmin"
+        }
+    ]
+}
```
trel, line 330: how does one set this in metadata?  i don't remember... might need an example.

docs/plugins/pluggable_rule_engine.md
korydraughn: That points to the code though. How would someone find the documentation for the plugin? Where would that live?

That also raises the question of - where are the docs for the C++ default REP? As I understand it, we don't provide any documentation for it right now.

docs/plugins/pluggable_rule_engine.md
korydraughn: They have to use `imeta` to attach the JSON as metadata.

docs/plugins/pluggable_rule_engine.md
trel: ah, i see.   correct on both points.

docs/plugins/pluggable_rule_engine.md
trel: oh, so ...

`imeta add -C /tempZone editors '["type": "group", "name": "rodsadmin"]'`

literally a JSON array as a string

trel: show me editors

trel: oh, just more JSON *in* the JSON...

trel: apparently an example using all the parts would be useful :P

docs/plugins/pluggable_rule_engine.md
korydraughn: Created https://github.com/irods/irods/issues/8290 for documenting the C++ default REP.

docs/plugins/pluggable_rule_engine.md
alanking: See line 334.

docs/plugins/pluggable_rule_engine.md
alanking: Linked to the new section.

docs/plugins/pluggable_rule_engine.md
alanking: So... do we want a more complete example for the `imeta` usage example? What do we want it to demonstrate?

trel: yes, a second example showing the additional use of the editors section would be helpful.

alanking: Added the example text to the `imeta` command and verified that it works.

alanking:
Squashed. Holding #s for now til https://github.com/irods/irods/pull/8289 is a little closer to merging.

Issue #327, opened 2025-04-01
Remove mentions of 3.x compatibility


alanking:
- [x] main only

Issue #307, opened 2025-02-13
Pluggable Network: Update information


alanking:
- [x] main
- [x] 4-3-stable

---

The information on the Pluggable Network page is out of date. Namely, `CS_NEG_DONT_CARE` is no longer the default in `acPreConnect` (it's `CS_NEG_REFUSE` now).

PR #329, opened 2025-04-02
Remove 3.x from documentation, update default acPreConnect value (main)

alanking:
Addresses #307
Addresses #327


alanking:
#'d, mergin

PR #338, opened 2025-04-16
iRODS 5 updates - configuration, process model, installation, genquery, etc. (main)

korydraughn:
There are a few TODOs scattered throughout the PR. Will address those soon, but feedback welcome.

TODOs:
- [x] Verify SQL in `GenQuery2 > Metadata Queries` section is correct
- [x] Update wording in `GenQuery2 > Escaping Bytes ...` to match that of the GenQuery1 section
- [x] Use either `resultset` or `result set` on GenQuery page, not both


docs/getting_started/installation.md
```diff
@@ -103,15 +103,15 @@ MariaDB can be used with the MySQL database plugin. Configuration is largely the
...
+    iRODS 4.3+ has a dependency on the Python 3 library, [pyodbc](https://pypi.org/project/pyodbc/). Setup will fail if this library is not installed.
 
 Installation of the iRODS server and PostgreSQL database plugin:
 
 ~~~
 $ (sudo) apt-get install irods-server irods-database-plugin-postgres
 ~~~
 
-On CentOS, some of the `irods-server` package dependencies can be satisfied from the EPEL repository provided by the `epel-release` package.
+On CentOS, some of the `irods-server` package dependencies can be satisfied from the EPEL repository provided by the `epel-release` package. TODO: Should we remove or update this line?
```
korydraughn, line 114: Handle TODO.

@SwooshyCueb Is this statement regarding epel-release still true for EL9? Can this line be removed now?

docs/system_overview/process_model.md
```diff
@@ -1,29 +1,36 @@
...
 
+| iRODS Process         | Expected Duration        | Expected Memory Usage           |
+| --------------------- | ------------------------ | ------------------------------- |
+| irodsServer           | Long-running             | ~34 MB                          |
+| irodsDelayServer      | Long-running             | ~33 MB                          |
+| irodsAgent (factory)  | Long-running             | ~92 MB                          |
+| irodsAgent            | Depends on client        | ~92 MB (minimum)                |
+
+TODO: Update or remove graphic.
+![iRODS Process Model Diagram](../images/process_model_diagram.jpg)
```
korydraughn, line 36: My initial thought is that the process model diagram can be removed.

@trel Thoughts?

docs/system_overview/genquery.md
```diff
@@ -841,14 +841,35 @@ iquery "select COLL_NAME, DATA_NAME where DATA_NAME = 'that''s my data.txt'"
...
 
-### Showing Duplicate Entries
+### Managing Duplicate/Unique Entries
 
-By default, GenQuery2 removes duplicate entries from the resultset.
+GenQuery2 no longer automatically inserts the DISTINCT keyword into the generated SQL. Users now have full control over the placement of the DISTINCT keyword.
 
-If you need to view duplicate entries, add `no distinct` after the `select` keyword. For example:
+!!! Important
+    The GenQuery2 implementation for iRODS 4.3.4 and earlier, originally removed duplicate entries from the resultset. This made it difficult for users to count records. Granting control over the DISTINCT keyword gives users more control and makes the behavior of the system more deterministic.
```
korydraughn, line 849: Add a statement about using GenQuery2 with earlier versions of iRODS.

docs/system_overview/configuration.md
```diff
@@ -508,6 +624,8 @@ An example `host_access_control` configuration:
 
 This is the main iRODS configuration file defining the iRODS environment. Any changes are effective immediately since iCommands reload their environment on every execution.
 
+The server no longer relies on information defined in this file.
```
korydraughn, line 627: Consider putting this in a `!!!` block.

docs/system_overview/configuration.md
```diff
@@ -12,7 +12,9 @@ This scrambled password file is saved after an `iinit` is run. If this file does
 
 ## /etc/irods/server_config.json
 
-This file defines the behavior of the server Agent that answers individual requests coming into iRODS. It is created and populated by the installer package.
+This file defines the behavior of the server-side processes that answer incoming requests coming into iRODS. It is created and populated by the installer package.
```
korydraughn, line 15: `the installer package` -> `setup_irods.py`

docs/system_overview/configuration.md
```diff
@@ -1,6 +1,6 @@
 #
 
-The following configuration files control nearly all aspects of how an iRODS deployment functions. All JSON files validate against the  [configuration schemas in each installation](https://github.com/irods/irods/tree/main/schemas/configuration/v4).
+The following configuration files control nearly all aspects of how an iRODS deployment functions. All JSON files validate against the  [configuration schemas in each installation](https://github.com/irods/irods/tree/main/schemas/configuration/v5).
```
korydraughn, line 3: `main` -> `5.0.0`

docs/getting_started/installation.md
```diff
@@ -128,24 +128,24 @@ The `setup_irods.py` script will ask for information in four (possibly five) sec
 
 2. Database Connection (if installing a 'provider')
     - ODBC Driver
-    - Database Server's Hostname or IP
+    - Database Server's FQDN, Hostname, or IP
```
korydraughn, line 131: Confirm this list covers everything.

docs/system_overview/configuration.md
```diff
@@ -12,7 +12,9 @@ This scrambled password file is saved after an `iinit` is run. If this file does
 
 ## /etc/irods/server_config.json
 
-This file defines the behavior of the server Agent that answers individual requests coming into iRODS. It is created and populated by the installer package.
+This file defines the behavior of the server-side processes that answer incoming requests coming into iRODS. It is created and populated by the installer package.
+
+iRODS 5 servers only require a working server_config.json file. All server properties which existed in the service account user's irods_environment.json file have been provided via the server_config.json file.
```
korydraughn, line 17: Rephrase this statement.

docs/system_overview/configuration.md
```diff
@@ -623,11 +741,25 @@ This is the main iRODS configuration file defining the iRODS environment. Any ch
...
+    // The following values are supported: none, cert, hostname
+    //
+    // When set to "none", authentication is skipped.
+    //
+    // When set to "cert", the server will verify that the certificate was signed by a trusted CA.
+    //
+    // When set to "hostname", the server will do everything defined by the "cert" level and then verify
+    // that the FQDN of the iRODS server matches either the common name or one of the subjectAltNames in
+    // the certificate.
+    "irods_ssl_verify_server": "hostname",
```
korydraughn, line 756: Confirm `hostname` is the actual default.

docs/system_overview/process_model.md
trel: i looked at it the other day - it doesn't seem ... wrong?     what about it is no longer true?

docs/system_overview/process_model.md
```diff
@@ -1,29 +1,36 @@
-#
+# iRODS Process Model
```
korydraughn, line 1: Change back to `#`

docs/system_overview/process_model.md
```diff
@@ -1,29 +1,36 @@
...
 
-iRODS 4.2+ launches a long-running server process (`/usr/sbin/irodsServer`), which immediately forks the Agent Factory and spawns the Delay Server (`/usr/sbin/irodsDelayServer`).  All incoming API requests are serviced by the Agent Factory, and for every new connection the Agent Factory spawns a new Agent.  Each Agent will service its request as quickly as possible, and then terminate.
+When iRODS starts, it launches a long-running server process:  
+`/usr/sbin/irodsServer`
```
korydraughn, line 4: Adjust this. It's weird when rendered.

docs/system_overview/process_model.md
```diff
@@ -1,29 +1,36 @@
...
+      │   ├─irodsAgent
+      │   └─irodsAgent
+      └─irodsDelayServer
 
+| iRODS Process         | Expected Duration        | Expected Memory Usage           |
+| --------------------- | ------------------------ | ------------------------------- |
+| irodsServer           | Long-running             | ~34 MB                          |
+| irodsDelayServer      | Long-running             | ~33 MB                          |
+| irodsAgent (factory)  | Long-running             | ~92 MB                          |
+| irodsAgent            | Depends on client        | ~92 MB (minimum)                |
```
korydraughn, line 33: Confirm memory usage values.

korydraughn, line 33: While the use of "Example" in the title isn't wrong, it feels out of place to me.

Try adding a statement which explains that the following structure is an example of what the process tree for iRODS could look like.

Could be helpful to include words about the table too.

docs/system_overview/configuration.md
```diff
@@ -434,6 +533,23 @@ Any changes made to the `server_config.json` file of a running iRODS server will
...
+    "advanced_settings": {
+        "dns_cache": {
+            "shared_memory_size_in_bytes": 5000000
+        },
+        "hostname_cache": {
+            "shared_memory_size_in_bytes": 2500000
+        }
+    }
+}
+```
```
korydraughn, line 549: The integer values for each property may lead to confusion. Perhaps it's better to list the properties as JSON pointers?
```
- /advanced_settings/dns_cache/shared_memory_size_in_bytes
- /advanced_settings/hostname_cache/shared_memory_size_in_bytes
```

docs/system_overview/process_model.md
korydraughn: The parts which mention `execAgent`, `socket filename`, `params` are no longer true.

The main reason is that I don't think it adds much value to the docs.

docs/system_overview/configuration.md
trel: do people use that enough?   what are you concerned is confusing about the actual JSON?

docs/system_overview/process_model.md
trel: if there are parts that are wrong/misleading - then certainly we don't want that.

i think with the text rendering just above... the image could be removed, yes.

docs/system_overview/process_model.md
trel: Example ....  could be replaced with ...

Standard
Representative
Common
Usual

and consider reordering the table so delayserver is last... like in the text tree

trel: and yes, a could sentences about the table would be good

docs/system_overview/configuration.md
trel: one reason to leave it as main might be that we tweak something after release, and then we don't have to remember to come back here and touch the link.

docs/system_overview/configuration.md
korydraughn: I'm looking at the values of the properties. They are just ... there.

Could replace those with something like `...` to signal that the values aren't the important bit.

docs/system_overview/configuration.md
korydraughn: True, but that also has the downside of making admins aware of properties their version doesn't support. They may be running iRODS 5, but looking at a schema for 6.0 without knowing it.

docs/system_overview/process_model.md
korydraughn: Agreed.

docs/system_overview/configuration.md
trel: oh.. so then we have a stable branch...   i was just hesitating on the very specific tag link that you had suggested.

docs/getting_started/installation.md
SwooshyCueb: I believe we still rely on epel on EL9.

trel: but mention of CentOS itself should be changed/removed.

docs/system_overview/configuration.md
trel, line 17:
```suggestion
iRODS 5 servers only require a working `server_config.json` file. All server properties which existed in prior versions of the service account's `irods_environment.json` file have been incorporated into `server_config.json`.
```

docs/system_overview/configuration.md
```diff
@@ -434,6 +533,23 @@ Any changes made to the `server_config.json` file of a running iRODS server will
 $ kill -HUP $(cat /var/run/irods/irods-server.pid)
 ```
 
+The following configuration properties require a full restart of the server to take effect.
```
trel, line 536:
```suggestion
Updates to the following configuration properties require a full restart of the server to take effect.
```

docs/system_overview/configuration.md
trel: oh i see... yes, you can probably just list them as pointers.   i think that's okay here.

are there only two?

docs/system_overview/genquery.md
```diff
@@ -841,14 +841,35 @@ iquery "select COLL_NAME, DATA_NAME where DATA_NAME = 'that''s my data.txt'"
 
 GenQuery2 will notice the use of double single quotes and collapse them to one single quote before sending to the database. In the case of the example, that means `that''s` will be passed to the database as `that's`.
 
-### Showing Duplicate Entries
+### Managing Duplicate/Unique Entries
 
-By default, GenQuery2 removes duplicate entries from the resultset.
+GenQuery2 no longer automatically inserts the DISTINCT keyword into the generated SQL. Users now have full control over the placement of the DISTINCT keyword.
```
trel, line 846:
```suggestion
GenQuery2 does not automatically insert the DISTINCT keyword into the generated SQL. Users have full control over the placement of the DISTINCT keyword.
```

docs/system_overview/genquery.md
trel, line 849:
```suggestion
    The initial GenQuery2 implementation for iRODS 4.3.4 and earlier removed duplicate entries from the resultset. This made it difficult for users to count records. Granting control over the DISTINCT keyword gives users more control and makes the behavior of the system more deterministic.
```

docs/system_overview/process_model.md
trel, line 4:
```suggestion
When iRODS starts, it launches the long-running server process `/usr/sbin/irodsServer`.
```

docs/system_overview/process_model.md
```diff
@@ -1,29 +1,36 @@
...
-       └─ irodsAgent
+## How requests are handled
 
-| iRODS Process         | Expected Duration        | Expected Resident Memory Usage  |
-| --------------------- | ------------------------ | ------------------------------- |
-| irodsServer           | long-running             | 126M                            |
-| irodsDelayServer      | long-running             | 17M                             |
-| irodsAgentFactory     | long-running             | 5M                              |
-| irodsAgent            | single client connection | 20M to 100M                     |
+The **Agent Factory** listens for new client connections. For each new connection, it forks a separate **Agent** process. Each Agent handles one client request, completes the work as quickly as possible, and then waits for the next request. Agents will continue to service requests until the client disconnects.
```
trel, line 13:
```suggestion
The **Agent Factory** listens for new client connections. For each new connection, it forks a separate **Agent** process. Each Agent handles the initial client request, completes the work as quickly as possible, and then waits for any additional requests. Agents will continue to service requests until the client disconnects.
```

docs/system_overview/process_model.md
```diff
@@ -1,29 +1,36 @@
...
 
-iRODS uses a dynamic linking model, rather than a static linking model, which adds significant overhead for short running processes (e.g. `ils`).  In order to mitigate this overhead, the irodsServer no longer calls `fork()` and `exec()` as a separate executable (pre-4.2).  In addition, since calling `fork()` within a threaded environment creates undefined behavior, the irodsServer calls `fork()` to spawn the Agent Factory process early, before any threading begins and any configuration is determined.
+## The Delay Server
 
-![iRODS Process Model Diagram](../images/process_model_diagram.jpg)
+The **Delay Server** is designed to remain idle most of the time. Every 30 seconds _(by default)_, it wakes up and spawns an Agent to check for any delayed rules that need to run. You can adjust this interval using the `delay_server_sleep_time_in_seconds` property in server_config.json. See [Configuration](../system_overview/configuration.md#etcirodsserver_configjson) for more information about this property.
```
trel, line 17: does it spawn an agent?  or does it make a client request which causes the agent factory to spawn an agent?

docs/system_overview/configuration.md
korydraughn: Those two options plus the access time options in R_GRID_CONFIGURATION.

So, there's a total of six options.

docs/system_overview/process_model.md
korydraughn: I think `spawns` is doing a lot of work in that sentence.

> or does it make a client request which causes the agent factory to spawn an agent?

Yes.

docs/system_overview/configuration.md
trel: okay, so maybe we make a section/statement that clearly talks about 'restart-required' configuration settings... and then we will have a full accounting in one place we can point to when someone inevitably asks.

docs/system_overview/process_model.md
trel: right, let's be targeted/specific in our words.   feels like it will help us in the long term.

docs/system_overview/configuration.md
korydraughn: Can also just drop the values, leaving just the property names.

With JavaScript highlighting. This is ok, but feels a little weird.
```js
{
    "advanced_settings": {
        "dns_cache": {
            "shared_memory_size_in_bytes"
        },
        "hostname_cache": {
            "shared_memory_size_in_bytes"
        }
    }
}
```

With JSON highlighting. This is likely to be closer to what we can expect when rendered. That is, the inner-most property is treated as a string instead of a property name.
```json
{
    "advanced_settings": {
        "dns_cache": {
            "shared_memory_size_in_bytes"
        },
        "hostname_cache": {
            "shared_memory_size_in_bytes"
        }
    }
}
```

And here's a version using `...`.
```js
{
    "advanced_settings": {
        "dns_cache": {
            "shared_memory_size_in_bytes": // ...
        },
        "hostname_cache": {
            "shared_memory_size_in_bytes": // ...
        }
    }
}
```

Or we go with the JSON pointer style and just mention the properties are using JSON pointer syntax. Can even link to documentation about JSON pointers for people who are unfamiliar with them.

korydraughn: > okay, so maybe we make a section/statement that clearly talks about 'restart-required' configuration settings... and then we will have a full accounting in one place we can point to when someone inevitably asks.

Sounds good.

trel: > Or we go with the JSON pointer style and just mention the properties are using JSON pointer syntax. Can even link to documentation about JSON pointers for people who are unfamiliar with them.

This feels the cleanest... having the JSON not be actual JSON looks weird/broken.

docs/system_overview/configuration.md
```diff
@@ -156,8 +169,24 @@ This file defines the behavior of the server Agent that answers individual reque
     // See "Checksum Configuration" for more details.
     "default_hash_scheme": "SHA256",
 
-    // Contains a set of key-value pairs of the form VARIABLE=VALUE such as "ORACLE_HOME=/full/path"
-    // from the server's environment. This setting can be empty.
+    // Encyption properties used for parallel transfer.
```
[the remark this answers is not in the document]
alanking: As I recall, if these are not properly configured, the client is not able to communicate with the server at all. In other words, I don't think this is only for parallel transfer encryption. I could be wrong.

docs/system_overview/configuration.md
```diff
@@ -391,9 +410,89 @@ This file defines the behavior of the server Agent that answers individual reque
...
+    // sockets using TCP keepalive. The range of acceptable values is defined by the range of acceptable values for the
+    // TCP_KEEPIDLE option. If a value outside of the acceptable range is provided, an error will be returned and the
+    // socket will be closed.
+    //
+    // -1 indicates that the socket option should remain unset.
+    "tcp_keepalive_time_in_seconds": -1,
+
+    // (Optional)
+    // Defines client-side TLS configurations. Although the "tls_client" object itself is optional,
+    // all sub-properties are required if "tls_client" is defined.
```
alanking, line 461: Looks like only "verify_server" is required. The other two are marked optional.

docs/system_overview/genquery.md
```diff
@@ -841,14 +841,35 @@ iquery "select COLL_NAME, DATA_NAME where DATA_NAME = 'that''s my data.txt'"
...
 
-If you need to view duplicate entries, add `no distinct` after the `select` keyword. For example:
+!!! Important
+    The GenQuery2 implementation for iRODS 4.3.4 and earlier, originally removed duplicate entries from the resultset. This made it difficult for users to count records. Granting control over the DISTINCT keyword gives users more control and makes the behavior of the system more deterministic.
+
+The following example now produces a resultset containing all replicas in the catalog.
```
alanking, line 851: Remove "now"?

docs/system_overview/process_model.md
alanking: "Expected"?

alanking (commented):
Very nice

docs/system_overview/configuration.md
korydraughn: Ok. Going with JSON pointers.

docs/getting_started/installation.md
korydraughn: Done.

docs/system_overview/process_model.md
korydraughn: Removed image.

docs/system_overview/genquery.md
korydraughn: Done.

docs/system_overview/configuration.md
korydraughn: Done.

docs/system_overview/configuration.md
korydraughn: Done.

docs/system_overview/process_model.md
korydraughn: Done.

docs/system_overview/process_model.md
korydraughn: Done.

docs/system_overview/process_model.md
korydraughn: Done.

korydraughn: Updated.

docs/system_overview/configuration.md
korydraughn: Done.

docs/system_overview/configuration.md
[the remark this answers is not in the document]
korydraughn: Done.

docs/system_overview/configuration.md
korydraughn: Done.

docs/system_overview/configuration.md
korydraughn: Done.

docs/system_overview/genquery.md
korydraughn: Done.

docs/system_overview/genquery.md
korydraughn: Done.

docs/system_overview/process_model.md
korydraughn: Done.

docs/system_overview/process_model.md
korydraughn: Done.

docs/system_overview/genquery.md
korydraughn: Done.

docs/system_overview/configuration.md
korydraughn: Fixed.

docs/system_overview/configuration.md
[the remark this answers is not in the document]
korydraughn: Those properties are required when using TLS. I don't know if they are required for legacy parallel transfer.

We're going to need to verify the requirements. I'm going to relax the wording for now and open an issue to revisit it. The docs can be updated post 5.0.

docs/system_overview/configuration.md
korydraughn: Out of scope for this PR. Will open an issue for this.

docs/system_overview/configuration.md
korydraughn: Leaving as is. We can adjust the links once 6.0 becomes a thing.

docs/system_overview/configuration.md
korydraughn: Did some debugging with @alanking and discovered the following:
- The TLS code treats 0 as an error and 1 as success
- `hostname` is an implicit default - i.e. the value is never explicitly set to `hostname`
- `cert` isn't a real value - i.e. any value other than `none`, `hostname`, or empty string results in `cert` behavior
  - The iRODS 5 schema limits the values to `none`, `hostname`, and `cert` which is good

We think `hostname` is indeed the default. I'm going adjust the wording to use `implicit`.

korydraughn: Will do.

korydraughn: Thinking about this a little more, the current wording is fine because iRODS 5 doesn't allow anything other than the three values.

The behavior should be documented in the server code.

docs/system_overview/configuration.md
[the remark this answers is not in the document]
korydraughn: Did a quick grep of the iRODS server and spotted use of the encryption algorithm in parallel transfer code and TLS code.

Decided to add TLS to the wording. Still going to open an issue so we confirm things.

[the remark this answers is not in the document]
korydraughn: Created #345.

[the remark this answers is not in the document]
korydraughn: Updated wording.

docs/system_overview/configuration.md
korydraughn: Done.

docs/getting_started/installation.md
korydraughn: Here's the current iRODS 5 setup prompts. Just posting these here so I can look them over later.

```
root@ee1fae0add9c:/# python3 /var/lib/irods/scripts/setup_irods.py 
iRODS server FQDN, hostname, or IP (253 characters max) [ee1fae0add9c]: 
Warning: Hostname `ee1fae0add9c` should be a fully qualified domain name.
iRODS user [irods]: 
iRODS group [irods]: 
iRODS server's role:
1. provider
2. consumer
Please select a number or choose 0 to enter a new value [1]: 
ODBC driver for postgres:
1. PostgreSQL ANSI
2. PostgreSQL Unicode
Please select a number or choose 0 to enter a new value [1]: 
Database server's hostname or IP address [localhost]: 
Database server's port [5432]: 
Database name [ICAT]: 
Database username [irods]: 

-------------------------------------------
Database Type: postgres
ODBC Driver:   PostgreSQL ANSI
Database Host: localhost
Database Port: 5432
Database Name: ICAT
Database User: irods
-------------------------------------------

Please confirm [yes]: 
Database password: 
Salt for passwords stored in the database: 
Local storage on this server [yes]: 
Default resource [demoResc]: 
iRODS Vault directory [/var/lib/irods/Vault]: 
iRODS server's zone name [tempZone]: 
iRODS server's port [1247]: 
iRODS port range (begin) [20000]: 
iRODS port range (end) [20199]: 
iRODS server's administrator username [rods]: 

-------------------------------------------
Zone name:                  tempZone
iRODS server port:          1247
iRODS port range (begin):   20000
iRODS port range (end):     20199
iRODS server administrator: rods
-------------------------------------------

Please confirm [yes]: 
iRODS server's zone key: 
iRODS server's negotiation key (32 characters): 

iRODS server's administrator password: 


=======================================================================

iRODS 5.0.0 tracks the access time for all replicas. Replicas which
existed before the upgrade will have an initial value of 0. This is
intentional, and the server will update access times as replicas
are opened.

If a starting value of 0 is unsuitable for your needs, consider setting
the access time to match the modification time using the following SQL:

    UPDATE R_DATA_MAIN
    SET access_ts = modify_ts
    WHERE access_ts = '0';

Please consult your local DBA before making any SQL adjustments.

=======================================================================
```

trel: consider expanding DBA to 'database administrator' for clarity.

korydraughn: Tweaked listing.

Opened PR to make prompts more consistent. See https://github.com/irods/irods/pull/8396.

korydraughn:
Code review comments have been resolved.

I think this is good enough for now.

korydraughn:
Squashed and pounded.

Tweaked some minor things. Those are below.

```
$ git diff origin/server_config.json.m 
```
```diff
diff --git a/docs/getting_started/installation.md b/docs/getting_started/installation.md
index d977db1..26156f2 100644
--- a/docs/getting_started/installation.md
+++ b/docs/getting_started/installation.md
@@ -103,7 +103,7 @@ MariaDB can be used with the MySQL database plugin. Configuration is largely the
 ### iRODS Setup
 
 !!! Important
-    iRODS 4.3+ has a dependency on the Python 3 library, [pyodbc](https://pypi.org/project/pyodbc/). Setup will fail if this library is not installed.
+    iRODS has a dependency on the Python 3 library, [pyodbc](https://pypi.org/project/pyodbc/). Setup will fail if this library is not installed.
 
 Installation of the iRODS server and PostgreSQL database plugin:
 
diff --git a/docs/system_overview/configuration.md b/docs/system_overview/configuration.md
index 6015d14..099f4dc 100644
--- a/docs/system_overview/configuration.md
+++ b/docs/system_overview/configuration.md
@@ -607,7 +607,7 @@ An example `host_access_control` configuration:
 This is the main iRODS configuration file defining the iRODS environment. Any changes are effective immediately since iCommands reload their environment on every execution.
 
 !!! Note
-    The server no longer relies on information defined in this file.
+    iRODS 5 does not rely on information defined in this file.
 
 ```json
 {
diff --git a/docs/system_overview/process_model.md b/docs/system_overview/process_model.md
index 57778bb..22da9cb 100644
--- a/docs/system_overview/process_model.md
+++ b/docs/system_overview/process_model.md
@@ -1,6 +1,6 @@
 #
 
-When iRODS starts, it launches the long-running server process, `irodsServer`.
+When iRODS starts, it launches the long-running server process `irodsServer`.
 
 This process immediately creates two other processes:

```

We'll do another pass over the docs to make sure they are good before the release.

PR #377, opened 2025-05-20
[379] update Doxyfile via Makefile

trel:


Makefile
```diff
@@ -20,9 +20,13 @@ default : doxygen mkdocs
 
 get_irods :
 	@echo "Getting iRODS source... v[${GIT_COMMITTISH}]"
-	@if [ -d ${IRODSTARGET} ] ; then rm -rf ${IRODSTARGET} ; fi
-	@git clone https://github.com/${GITHUB_ACCOUNT}/irods ${IRODSTARGET}
+	@if [ ! -d ${IRODSTARGET} ] ; then \
+		git clone https://github.com/${GITHUB_ACCOUNT}/irods ${IRODSTARGET}; \
+		else \
+		echo "IRODSTARGET already found ... ${IRODSTARGET}"; \
+		fi
```
korydraughn, line 27: This change might conflict with c066080a07ef02e45ee7f6bc9c3282399c1f072d.

trel: it shouldn't... that code just guarantees the file will be there... (well, technically just its parent directory, but sheesh, don't be moving these things around yourself)...

and then the `sed` will just manipulate the one line in that file.

trel:
this updates the PROJECT_NUMBER to be consistent with the GIT_COMMITTISH

Makefile
trel: oh, reading again...

korydraughn: The original code solved a situation in which a dev is updating documentation in irods/irods, pushing changes to their remote branch, and then rebuilding the docs. It also solved switching branches.

Your change would remove that ability.

trel: hmm, that is true.

the goal of this change was to be able to edit the version getting shown in the header of the rendered docs.

maybe the auto-updating with the sed line is enough...   it will always get deleted/cloned again... but then immediately be updated to be the same as GIT_COMMITTISH

korydraughn: You can add a new option that enables what you want. Then it becomes a matter of passing the correct options.

trel: added a new commit.  resolving.

trel:
okay, now, just using the GIT_COMMITTISH after the always-new-clone.

confirmed it's working in both scenarios, and respects command line arguments coming into the docker command, as documented.

alanking:
Can we make an issue specific to what this is trying to address other than the housekeeping issue?

trel:
yes, will do.

trel:
renumbered

alanking (commented):
I say squash

trel:
squashed with some words

alanking (approved):
\# it

trel:
#'d

PR #383, opened 2025-07-03
[irods/irods#6826] Federation: Describe consumer redirection

alanking:
In service of irods/irods#6826

Not sure what else to write about this but definitely open to suggestions on what we might like to see here.


korydraughn (approved):
Great! Ship it.

alanking:
#'d, mergin

PR #388, opened 2025-09-09
[irods/irods#8605] Document Consortium's stance on API and ABI stability (main)

korydraughn:
Here's a screenshot of the rendered documentation.

image

And here's what the navigation menu looks like. Tried to remove the button to the left of the section title, but failed. Also didn't spend a lot of time on that part.

image

### DIFF
--- a/docs/developers/api_abi_stability.md
+++ b/docs/developers/api_abi_stability.md
@@ -1,0 +1,5 @@
+# API and ABI Stability
+
+The iRODS Consortium strives to maintain API stability and not break backward compatibility within a major version. This statement applies to stable public APIs only. Experimental features and projects which have not reached 1.0 status may experience breaking changes. Consumers of these features and projects must keep that in mind as they consider deploying them into production.
+
+The iRODS Consortium DOES NOT make any guarantees around ABI stability. This applies to iRODS libraries written in C, C++, and other languages which compile to machine code.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -77,6 +77,7 @@ nav:
     - Doxygen: 'doxygen/index.html'
     - Contributing: 'developers/contributing.md'
     - Development Tools: 'developers/development_tools.md'
+    - API and ABI Stability: 'developers/api_abi_stability.md'
     - Library Examples: 'developers/library_examples.md'
     - Logging: 'developers/logging.md'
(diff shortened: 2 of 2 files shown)


alanking (commented):
Looks good to me.

My only hesitation is whether this needs its own dedicated page. I'm wondering whether there's a more general space we can use or create. Maybe a slightly more general "Version Compatibility" page or something like that? It feels like it's very limited in its scope for what could go in here in the future. All that being said, I'm fine with things as they are as well.

korydraughn:
That crossed my mind as well. If you see an alternative place to put it, let me know.

Given your recent issue about `char*` and `const char*`, I'm now starting to reconsider whether ABI stability should be a thing we maintain. It can be hard to do which is why I've tried to avoid it.

trel:
agreed - let's hold on this for a little bit and think through what we DO want to say... as well as whether there is a more general place for it than its own page.

korydraughn:
Previous/Latest Notes:
- Minor releases maintain API compabitlity
- Patch releases strive for ABI compatibility

korydraughn:
Updated wording and moved it to the contributing page of the developers section.

korydraughn:
The rendered text.

image

docs/developers/contributing.md
```diff
@@ -73,6 +73,28 @@ Then, attend to any checkboxes and issues that may need to be closed.  If you do
...
+
+To help in achieving that goal, the following rules must be honored at all times:
+
+- No modification or removal of packing instructions
+- No modification or removal of data types used in public APIs
+- No modification or removal of data types sent over the network
+
+If modifications are needed, the developer must create new APIs to support it.
+
+These rules DO NOT apply to experimental APIs.
```
trel: consider...

```suggestion
These rules DO NOT apply to APIs designated as `experimental`.
```

trel (commented):
i like it.

docs/developers/contributing.md
korydraughn: Done.

Dropped the backticks.

trel:
squash em

korydraughn:
Squashed.

trel (approved):
pound it

korydraughn:
Added pound.

Merging.

PR #390, opened 2025-09-16
[irods/irods#8651] Add release notes for 5.0.2 (main)

korydraughn:
A couple updates regarding the release notes ...
- They no longer report the total number of closed issues
- They no longer report information across branches

The decision to leave that info out was due to the amount of effort needed to obtain it and how much value it provides to the reader.

The release notes highlight the important bits. If including counts, etc. in the notes is important, then we'll need to develop a scheme which makes it easy to obtain those numbers easily and with confidence.

### DIFF
--- a/docs/release_notes.md
+++ b/docs/release_notes.md
@@ -1,5 +1,57 @@
 #
 
+## 5.0.2
+
+Release Date: 2025-10-01
+
+The iRODS Consortium and RENCI are pleased to announce iRODS 5.0.2.
+
+This release adds support for Enterprise Linux 10 and Debian 13. With that comes bug fixes and improvements for `irsync`, GenQuery2, unattended installation, physical quotas, etc.
+
+Important updates regarding the plugins:
+
+- Plugin version numbers now follow three segments instead of four, aligning with semantic versioning (X.Y.Z instead of X.Y.Z.W).
+- Plugin versioning is independent of the server version. Keep in mind that package names include the version of the development library used during the build.
+
+The latest binary packages for Enterprise Linux 9, Enterprise Linux 10, Ubuntu 22.04, Ubuntu 24.04, Debian 12, and Debian 13 are available at <https://packages.irods.org/>.
+
+### Changed
+
+- Update help text for `itouch` and `irepl` to be more explicit (#4687, #8454, #8673).
+- Do not print stacktrace when `irods_environment.json` file is not readable (#8083).
+- Document SIGHUP signal in server binary's help text (#8582).
+- Invoke `fileNotify` when closing replica (#8586, irods/irods_resource_plugin_s3#2146).
+- Detect zombie processes during shutdown and warn the administrator (#8588).
+- Create stored procedures for MySQL using ODBC directly (#8601, #8686).
+- Include access time grid configuration options in help text for `iadmin` (#8630).
+
+### Fixed
+
+- Fix renaming of collections containing multi-byte characters (#6239).
+- Maintain appropriate permissions for `msiExecCmd_bin` directory on package upgrade (#6671).
+- Do not create target collections or directories during dry-run of `irsync` (#7774).
+- Fix `irsync`'s handling of `ALL_KW` when targeting a single data object (#8295).
+- Limit max value of `maximum_size_for_single_buffer_in_megabytes` for DataObjGet API (#8373).
+- Improve `irsync`'s handling for checksum errors (#8384).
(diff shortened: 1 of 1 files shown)


docs/release_notes.md
```diff
@@ -1,5 +1,43 @@
 #
 
+## 5.0.2
+
+Release Date: 2025-09-XX
```
korydraughn, line 5: Update once release testing is completed.

docs/release_notes.md
```diff
@@ -1,5 +1,43 @@
...
+
+The iRODS Consortium and RENCI are pleased to announce iRODS 5.0.2.
+
+This patch release adds support for Enterprise Linux 10 and Debian 13. With that comes bug fixes and improvements for `irsync`, GenQuery2, unattended installation, physical quotas, etc.
+
+This release consists of [30 commits from 9 contributors](https://github.com/irods/irods/compare/5.0.1...5.0.2).
+
+The latest binary packages for RockyLinux9, RockyLinux10, Ubuntu22, Ubuntu24, Debian12, and Debian13 are available at <https://packages.irods.org/>.
```
korydraughn, line 13: Hmm, rockylinux or enterprise linux?

korydraughn:
I agree it seems correct to include the milestone, but it doesn't cover 4.3.5 issues and is incorrectly listed as the full github listing. That's the only reason why I didn't include it. I couldn't figure out how to construct a URL which covered 4.3.5 issues. Thoughts?

docs/release_notes.md
korydraughn: For release testing, we'll lock in the OS for each version of EL. For example, almalinux 9 for EL9 and rockylinux 10 for EL10.

@SwooshyCueb Any thoughts on this conversation?

trel:
You'd have to do two links like we have in the past...

docs/release_notes.md
trel: 9 has only been rocky i believe.

korydraughn:
The problem with including info about 4.3.5 is that not all commits against 4.3.5 are included in 5.0.2. GitHub doesn't provide a way to obtain the list of issues which are part of both branches.

Not sure how we can address that, but will continue to think about it.

korydraughn:
Here's the issue containing references to many PRs which handle it.
- https://github.com/irods/irods/issues/6008

Here's two diffs showing what you want.
- https://github.com/irods/irods/commit/5e3451915688d6361359556154c80e3fb2db0048
- https://github.com/irods/irods_development_environment/commit/19839ebafea90259867715678ced1eb3ccfa5051

korydraughn:
@SwooshyCueb Is this the replacement for `apt-key`?

```
mkdir -p /etc/apt/keyrings && \
    wget -qO - https://packages.irods.org/irods-signing-key.asc | \
        gpg \
            --no-options \
            --no-default-keyring \
            --no-auto-check-trustdb \
            --homedir /dev/null \
            --no-keyring \
            --import-options import-export \
            --output /etc/apt/keyrings/renci-irods-archive-keyring.pgp \
            --import \
        && \
    echo "deb [signed-by=/etc/apt/keyrings/renci-irods-archive-keyring.pgp arch=amd64] https://packages.irods.org/apt/ $(lsb_release -sc) main" | \
        tee /etc/apt/sources.list.d/renci-irods.list
```

trel:
I was hoping for the simpler...
```
wget -qO- https://packages.irods.org/irods-signing-key.asc | sudo tee /etc/apt/trusted.gpg.d/renci-irods.asc
```
or
```
wget -qO- https://packages.irods.org/irods-signing-key.asc | sudo tee /etc/apt/keyrings/renci-irods.asc
```

SwooshyCueb:
While some versions of apt will take the asc file as-is, we want to use a format that can contain multiple keys.

Let's split them into the individual commands
```
mkdir -p /etc/apt/keyrings
wget -qO - https://packages.irods.org/irods-signing-key.asc | \
    gpg \
        --no-options \
        --no-default-keyring \
        --no-auto-check-trustdb \
        --homedir /dev/null \
        --no-keyring \
        --import-options import-export \
        --output /etc/apt/keyrings/renci-irods-archive-keyring.pgp \
        --import
echo "deb [signed-by=/etc/apt/keyrings/renci-irods-archive-keyring.pgp arch=amd64] https://packages.irods.org/apt/ $(lsb_release -sc) main" | \
     tee /etc/apt/sources.list.d/renci-irods.list
```

Issue #417, opened 2026-04-29
Documentation should indicate against relying on case-insensitivity when collating DB results


d-w-moore:
Related: https://github.com/irods/python-irodsclient/issues/600
Add to GenQuery 1 & 2 and/or specific query related  docs( as well as in any more general DB / query related sections that might exist ) a general caution against depending on lexicographical order in result sets from any query in iRODS.

This is because:

In default installations of MySQL, `select 1 where 'a' = 'A'` yields a row  containing 1 (ie, the condition  is true). However, PostgreSQL considers `'A' > 'a'` as being true, whereas SQLite3 evaluates `'a' > 'A'` as true.

This will affect, at the very least, any of the following:
   - sorting in ORDER BY 
   - result sets depending on  > , = and < and similar operators  when comparing strings
   - result sets depending on `BETWEEN` string comparisons.

d-w-moore:
This seems to affect GenQuery1 and would affect any query interface that is simply a pass-thru to the database for collative  operators such as `>` and `BETWEEN` (as opposed to matching operators, such as `LIKE` or `=`). Consider if data objects named 'a', 'A' and 'a_' are defined (and for this simple case, no other data objects).

Then under PostgreSQL defaults:
```
$ iquest "select  DATA_NAME where DATA_NAME between 'a' 'a_'"
DATA_NAME = a
------------------------------------------------------------
DATA_NAME = a_
------------------------------------------------------------

```

but under MySQL defaults:
```
$ iquest "select  DATA_NAME where DATA_NAME between 'a' 'a_'"
DATA_NAME = a
------------------------------------------------------------
DATA_NAME = A
------------------------------------------------------------
DATA_NAME = a_
------------------------------------------------------------
```

This is consistent with the MySQL property of considering 'a' = 'A' to be true in MySQL, as noted in the description above .  Although ,notably, given the aforementioned existing data objects, case-sensitive defaults for **MySQL** still do not consider them equal in matching contexts:
```
$ iquest "select  DATA_NAME where DATA_NAME = 'a'"
DATA_NAME = a
------------------------------------------------------------
$ iquest "select  DATA_NAME where DATA_NAME = 'A'"
DATA_NAME = A
------------------------------------------------------------
``` 

[...]

PR #420, opened 2026-05-05
[#417] warn against assumed collation order

d-w-moore:


trel:
i think this covers the concern.

and yes, separate section, please.

docs/system_overview/genquery.md
```diff
@@ -201,6 +201,21 @@ There are a few other options that can be used with GenQuery to affect how the r
  - no-distinct: Instructs GenQuery to return all results, even repeating identical results where applicable.
  - uppercase: When specified, all VALUEs should be supplied in UPPERCASE and the query will be made case-insensitive.
 
+### Collation Order
+
+It should be noted, with regard to the case-sensitive query defaults, that one cannot always rely on an assumed
+collation order (i.e. the result of comparing mixed-case string arguments) to be followed.  Because GenQuery passes
```
korydraughn:
```suggestion
collation order (i.e. the result of comparing mixed-case string arguments) to be followed. Because GenQuery passes
```

docs/system_overview/genquery.md
```diff
@@ -201,6 +201,21 @@ There are a few other options that can be used with GenQuery to affect how the r
...
+operators such as <, =, >, BETWEEN, and ORDER directly to the backing database, such assumptions are inherently
+non-portable.  Consider the following query, when run in the context of three data objects named `a`, `A`, and `a_`:
```
korydraughn:
```suggestion
non-portable. Consider the following query, when run in the context of three data objects named `a`, `A`, and `a_`:
```

docs/system_overview/genquery.md
```diff
@@ -201,6 +201,21 @@ There are a few other options that can be used with GenQuery to affect how the r
...
+
+```sql
+select DATA_NAME where DATA_NAME between 'a' 'a_'
+```
+
+Under the default American setup of MySQL, for example, 'A' will fall in the range defined by `between` (as corroborated
```
korydraughn:
```suggestion
Under the default American setup of MySQL, for example, `A` will fall in the range defined by `between` (as corroborated 
```

docs/system_overview/genquery.md
korydraughn: To maintain consistency throughout the file, please combine these lines so that they are one long line. The rendered text should not be affected by the change.

docs/system_overview/genquery.md
```diff
@@ -201,6 +201,21 @@ There are a few other options that can be used with GenQuery to affect how the r
...
+operators such as <, =, >, BETWEEN, and ORDER directly to the backing database, such assumptions are inherently
+non-portable.  Consider the following query, when run in the context of three data objects named `a`, `A`, and `a_`:
+
+```sql
+select DATA_NAME where DATA_NAME between 'a' 'a_'
+```
+
+Under the default American setup of MySQL, for example, 'A' will fall in the range defined by `between` (as corroborated 
+by testing the result of the query `select 'a' <= 'A' and 'A' <= 'a_'` in the mysql client) and thus be reported among
+GenQuery's results; whereas under the default setup of PostgreSQL, it will not.
```
korydraughn: To maintain consistency throughout the file, please combine these lines so that they are one long line. The rendered text should not be affected by the change.

docs/system_overview/genquery.md
```diff
@@ -201,6 +201,21 @@ There are a few other options that can be used with GenQuery to affect how the r
  - no-distinct: Instructs GenQuery to return all results, even repeating identical results where applicable.
  - uppercase: When specified, all VALUEs should be supplied in UPPERCASE and the query will be made case-insensitive.
 
+### Collation Order
+
+It should be noted, with regard to the case-sensitive query defaults, that one cannot always rely on an assumed
+collation order (i.e. the result of comparing mixed-case string arguments) to be followed.  Because GenQuery passes
+operators such as <, =, >, BETWEEN, and ORDER directly to the backing database, such assumptions are inherently
```
korydraughn: Should `ORDER` be `ORDER-BY`?

docs/system_overview/genquery.md
alanking: Are these lines being manually wrapped on purpose? Just noticing that all the text around this section does not split up lines like this - not sure how that affects appearance in the page rendering.

d-w-moore:
> i think this covers the concern.
> 
> and yes, separate section, please.

Still thinking about the wording though - what do you think about: 
```

Such assumptions are impacted by the iRODS administrator's selection and configuration of the backing database
which, in turn, implements GenQuery's  <, =, >, BETWEEN, and ORDER BY operators.
```

korydraughn (commented):
Good stuff.

I was going to recommend moving this into a third section for **shared** behavior, but that isn't necessary at this time.

alanking (commented):
Seems good. Just noticed one thing

docs/system_overview/genquery.md
alanking: Oops, looks like @korydraughn found this first. Resolving.

docs/system_overview/genquery.md
d-w-moore: yes . except in GEnQuery it's just ORDER and ORDER_DESC

korydraughn: Oh right. Forgot about that in GenQuery1.

docs/system_overview/genquery.md
```diff
@@ -201,6 +201,18 @@ There are a few other options that can be used with GenQuery to affect how the r
...
+
+It should be noted, with regard to the case-sensitive query defaults, that one cannot always rely on an assumed collation order (i.e. the result of comparing mixed-case string arguments) to be followed.  That is because a particular de facto ordering within a character set - as implied by GenQuery operators such as <, =, >, BETWEEN, and ORDER (BY) - can easily be impacted by the configuration and design of the backing DBRMS which implements those operators.
+
+Consider the following query for example, when run in the context of three data objects named `a`, `A`, and `a_`:
+
+```sql
+select DATA_NAME where DATA_NAME between 'a' 'a_'
+```
+
+Under the default American setup of MySQL, for example, 'A' will fall in the range defined by `between` (as corroborated by testing the result of the query `select 'a' <= 'A' and 'A' <= 'a_'` in the mysql client) and thus be reported among GenQuery's results; whereas under the default setup of PostgreSQL, it will not.
```
korydraughn, line 214: Is "American" how MySQL refers to the default setup?

d-w-moore: No, although I'd assume it to depend on nationality almost certainly. 

Maybe United States  would be better.  Now that I think on it.

d-w-moore: "default MySQL settings for English language  installations in the U.S."?

korydraughn: I recommend looking at the official MySQL docs to see how they describe it. Alternatively, you can mention the options and values which directly influence the behavior. Mentioning the charset or collation options, etc. will make things very clear.

d-w-moore: ok. will do

